PR #19, opened 2022-11-21
Update uniswap/compute_price

ColinPR:
Small mistake in first condition. Price reads correct for both tokens now.


0xJepsen (approved):
lgtm

Issue #25, opened 2022-11-24
Fixing wsteth, wseth naming in tokens/mod.rs


eljhfx:
Lines 140 and 141 call it wseth and everywhere else, it's wsteth. Feels like that was a mistake. Wdyt?

![Screen Shot 2022-11-23 at 8 43 11 PM](https://user-images.githubusercontent.com/56851203/203675448-5c61bcd2-0639-4663-8a00-703767f1cc4c.png)

_Originally posted by @eljhfx in https://github.com/primitivefinance/arbiter/discussions/24_

eljhfx:
@0xJepsen happy to take this one sir

eljhfx:
https://github.com/primitivefinance/arbiter/pull/28

Autoparallel:
Closing issue. Newest updates in `main` branch have this fixed now.

PR #28, opened 2022-11-29
fixing wseth -> wsteth

eljhfx:
Issue #25 

Just changed wseth to wsteth, where it should be (small change lol)


ColinPR (approved):
Looks great. Thanks for the fix!

PR #31, opened 2022-12-23
Change token builder to pub

Autoparallel:
Small fix. Just double check I'm not crazy here.

### DIFF
--- a/app/src/tokens/mod.rs
+++ b/app/src/tokens/mod.rs
@@ -16,7 +16,7 @@ pub struct Token {
 
 impl Token {
     /// Public builder function that instantiates a `Token`.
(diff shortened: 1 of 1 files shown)


app/src/tokens/mod.rs
```diff
@@ -16,7 +16,7 @@ pub struct Token {
 
 impl Token {
     /// Public builder function that instantiates a `Token`.
-    fn new(address: H160, decimals: u16, name: String, is_stable: bool) -> Self {
+    pub fn new(address: H160, decimals: u16, name: String, is_stable: bool) -> Self {
```
0xJepsen, line 19: You're not crazy, we do need this

0xJepsen (approved):
lgtm

Issue #33, opened 2022-12-23
Add a config.yaml for input arguments


Autoparallel:
Right now, we have CLI arguments and we have environment arguments. I think that it would be best to implement inputs in a uniform way that is also a good experience.

I think the best way around this is to use CLI arguments with defaults but allow the user to load a `config.yaml` file. This would be the same set up as [Erigon](https://github.com/ledgerwatch/erigon). 

The main tasks are then:
- [x] Allow for a flag like `--config-file` where a user can provide a `config.yaml`. 
- [x] Set up a preconfigured `config.yaml` file.
- [x] Change API/RPC away from environment variables and into CLI configs.

---

What do you think? @0xJepsen  @eljhfx @ts0yu

Autoparallel:
@0xJepsen Do you prefer a `config.toml`?

ts0yu:
a .toml file might be a bit nicer

0xJepsen:
Let us do a toml then.

ts0yu:
Task 1 is done https://github.com/primitivefinance/arbiter/pull/43

PR #43, opened 2023-01-01
cli with clap derive

ts0yu:
- clap derive feature so our cli is more expandable and modular
- `config.toml` arg for later parsing, see https://github.com/primitivefinance/arbiter/issues/33


0xJepsen (changes requested):
This looks good but warrants some changes to the readme. I think we should eventually change the default run behavior at some point, but for now, let's keep it until we have a good idea of what we want.

ts0yu:
made the requested changes

0xJepsen (approved):
This looks good to me!

PR #47, opened 2023-01-07
Toml

### LINKED ISSUE #33: Add a config.yaml for input arguments

0xJepsen:
This PR adds support for a configuration file and closes #33 
the PR includes additions to the README on how to use the configuration file.


README.md
```diff
@@ -90,22 +105,35 @@ Liquidity: 21972098821216706277
 Tick:      205282
 Price:     "1.216568804789000000000000000000000000000e+3"
 ```
+
 You may also build the executable with `cargo build`, which will output a binary in `target/`
 
+## Configuration File
+
+To run arbiter with a configuration file, we can run the command `cargo run see --config true` or, if we have built the latest changes, `arbiter see --config true`. This will load in configuration settings from a `config.toml` file located in `./crates/cli/src/`. Feel free to make custom configuration files for your needs. The logic that finds the file is in `crates/cli/src/config/mod.rs`; remember to add the path of your new file to the list of file paths.
```
ts0yu, line 113: We need to update this so it reflects the latest changes of being able to run `arbiter see --config`, without the true parameter.

crates/clairvoyance/src/uniswap.rs
```diff
@@ -88,6 +92,26 @@ impl Pool {
...
 }
 
+// Creates a pool from the cli/or config parameters
+pub async fn get_pool(
+    token0: &String,
+    token1: &String,
+    bp: &str,
+    provider: Arc<Provider<Http>>,
+) -> Result<Pool, Box<dyn Error>> {
+    // let provider = match env::var_os("PROVIDER") {
```
ts0yu, line 102: let's remove the comments before we merge with main

crates/cli/src/config/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                see: None,
+            }
+        });
+        // ! check the rpc_url field in the network config.toml section
+        let rpc_url = match config_toml.network {
+            Some(network) => network.rpc_url.unwrap_or_else(|| {
+                println!("Missing Field rpc_url in network config");
+                "unknown".to_owned()
+            }),
+            None => "unknown".to_owned(),
```
ts0yu, line 63: Setting the field to `"unknown"` is basically just letting the program panic at a later stage so its basically redundant here imo

crates/cli/src/config/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+        };
+        // ! check the token0, token1, and bp fields in the see config.toml section
+        let (token0, token1, bp) = match config_toml.see {
+            Some(see) => {
+                let token0 = see.token0.unwrap_or_else(|| {
+                    println!("Missing Field token0 in see config");
+                    "unknown".to_owned()
```
ts0yu, line 70: same here and below

crates/cli/src/main.rs
```diff
@@ -62,31 +60,37 @@ async fn main() -> Result<()> {
...
-            for pool in pools {
-                join!(pool.monitor_pool());
+            match config {
+                true => {
+                    // If present, load config.toml and get pool from there
+                    println!("Loading config.toml...");
+                    let config_obj = config::Config::new();
+                    println!("Config.toml loaded successfully! Getting Pool...");
+                    let pool = get_pool(
+                        &config_obj.token0,
```
ts0yu, line 70: the load will appear to be successful when in fact `token0` is `"unknown"`

ts0yu (changes requested):
looking good! just a few concerns about the error handling, but otherwise solid stuff.
i can make some changes for custom error handling with `thiserror` tomorrow, if you want

crates/cli/src/config/mod.rs
0xJepsen: This would be pretty good imo

crates/cli/src/main.rs
0xJepsen: Good point, i replaced with a more correct statement

ts0yu (approved):

PR #49, opened 2023-01-08
custom config error type

ts0yu:
adds the error type + better propegation to config we discussed


0xJepsen (approved):
This is clean. LGTM

PR #52, opened 2023-01-08
reduce code repetition in cli

ts0yu:


0xJepsen (approved):
Nice, this is a little cleaner and will scale well as we add other features.

PR #54, opened 2023-01-09
Update token.rs

Autoparallel:
Added RPL token.

Please double check the data for the token! 

https://etherscan.io/token/0xd33526068d116ce69f19a9ee46f0bd304f21a51f


0xJepsen (approved):
Double checked, looks good to me. If you are ever unsure, you can check the ERC contract on the etherscan address.

Autoparallel:
Deleted the branch since the change was so small.

PR #55, opened 2023-01-10
GBM pricing

Autoparallel:
Basic implementation done. A lot more needs to be done but we have the right parameters and we can output a plot! 

I will create more issues later.


0xJepsen (approved):
This looks like a great evolvement, could be cleaned up a bit in the future but this looks like a great starting point!

PR #58, opened 2023-01-10
Price impact features

### LINKED ISSUE #57: In the uniswap pool structure create mutable traits that update with monitor pool
liquidity and tick are the obvious ones right now...

0xJepsen:
This closes #57

### DIFF
--- a/crates/clairvoyance/src/uniswap.rs
+++ b/crates/clairvoyance/src/uniswap.rs
@@ -19,17 +19,21 @@ use std::error::Error;
 #[derive(Debug, Clone)]
 pub struct Pool {
     /// Token 0.
-    pub token_0: Token,
+    token_0: Token,
     /// Token 1.
-    pub token_1: Token,
+    token_1: Token,
     /// Basis point of pool.
-    pub bp: u32,
+    bp: u32,
     /// Address of the pool.
-    pub address: H160,
+    address: H160,
     /// Factory that created the pool. This could be generic in future.
-    pub factory: UniswapV3Factory<Provider<Http>>,
+    factory: UniswapV3Factory<Provider<Http>>,
     /// Pool contract object.
-    pub inner: IUniswapV3Pool<Provider<Http>>,
+    inner: IUniswapV3Pool<Provider<Http>>,
+    /// Current Tick.
+    tick: i32,
+    /// Current liquidity.
+    liquidity: u128,
 }
 
 impl Pool {
@@ -39,7 +43,7 @@ impl Pool {
         token_1: Token,
         bp: u32,
         provider: Arc<Provider<Http>>,
-    ) -> Result<Self, ()> {
+    ) -> Result<Pool, ()> {
         match bp {
             1 | 5 | 30 | 100 => (),
             _ => return Err(()),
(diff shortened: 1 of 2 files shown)


crates/clairvoyance/src/uniswap.rs
```diff
@@ -15,68 +15,129 @@ use bindings::uniswap_v3_factory::UniswapV3Factory;
...
+    /// Current Tick.
+    tick: i32,
+    /// Current liquidity.
+    liquidity: u128,
+    /// sqrt_price_x96
+    sqrt_price_x96: ethers::types::U256,
 }
 
 impl Pool {
+    /// Getters and setters for error checking and access controls.
```
ts0yu, line 43: no point in writing getter and setter functions when we can just make the struct field public...

crates/clairvoyance/src/uniswap.rs
```diff
@@ -15,68 +15,129 @@ use bindings::uniswap_v3_factory::UniswapV3Factory;
...
-        println!("Got Pool: {:#?}. Listening for events...", pool.address());
-        let tokens = (self.token_0.clone(), self.token_1.clone());
-        let swap_events = pool.swap_filter();
-        let pool_token_0 = pool.token_0().call().await.unwrap();
-        let mut swap_stream = swap_events.stream().await.unwrap();
+    pub async fn monitor_pool(&mut self) {
 
+        let pool_contract = self.get_pool_contract();
+        println!("Got Pool: {:#?}. Listening for events...", pool_contract.address());
+        let pool_tokens = self.get_pool_tokens();
```
ts0yu, line 130: same things with getters and setters here. we can just expose the struct field publicly and access it here

crates/clairvoyance/src/uniswap.rs
```diff
@@ -15,68 +15,129 @@ use bindings::uniswap_v3_factory::UniswapV3Factory;
...
+        let swap_events = pool_contract.swap_filter();
+        let pool_token_0 = pool_contract.token_0().call().await.unwrap();
+        let mut swap_stream = swap_events.stream().await.unwrap();
         while let Some(Ok(event)) = swap_stream.next().await {
+            let (tick, liq, sqrtprice) = (event.tick, event.liquidity, event.sqrt_price_x96);
+            self.set_pool_tick(tick);
```
ts0yu, line 136: same here

crates/clairvoyance/src/uniswap.rs
```diff
@@ -85,13 +146,26 @@ impl Pool {
...
-            )
+                compute_price(pool_tokens.clone(), event.sqrt_price_x96, pool_token_0,).to_string()
+            );
+            // Check tick, price, and liquidity where updated
+            assert_eq!(event.tick, self.get_pool_tick());
+            assert_eq!(event.liquidity, self.get_pool_liquidity());
+            assert_eq!(event.sqrt_price_x96, self.get_sqrt_price_x96());
         }
     }
+    // Psudo Code for price impact work don't remove yet
```
ts0yu, line 157: lets remove this if we are pushing to main

crates/cli/src/main.rs
```diff
@@ -107,12 +107,12 @@ async fn main() -> Result<()> {
...
 
                     // Get pool from CLI/defaults.
-                    let pool: Pool = get_pool(token0, token1, bp, provider).await.unwrap();
+                    let pool = get_pool(token0, token1, bp, provider).await.unwrap();
 
                     vec![pool]
                 }
             };
-            for pool in pools {
+            for mut pool in pools {
```
ts0yu, line 115: why does this need to be mutable?

ts0yu (dismissed):

crates/clairvoyance/src/uniswap.rs
0xJepsen: yeah but we don't want to make them public because we want to ensure they are always correct and thus only changed by functions with the correct logic.

crates/clairvoyance/src/uniswap.rs
```diff
@@ -15,68 +15,129 @@ use bindings::uniswap_v3_factory::UniswapV3Factory;
...
+        self.tick = tick;
+    }
+    fn set_pool_liquidity(&mut self, liquidity: u128) {
+        self.liquidity = liquidity;
+    }
+    fn set_pool_sqrt_price_x96(&mut self, sqrt_price_x96: ethers::types::U256) {
+        self.sqrt_price_x96 = sqrt_price_x96;
+    }
+    /// Updates the pool tick and liquidity manually with a contract call.
+    pub async fn _update_pool_tick_and_price(&mut self) {
```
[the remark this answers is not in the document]
0xJepsen: This one is how we would update it correctly if we wanted to. This makes sure it will always be updated to the correct contract state.

crates/clairvoyance/src/uniswap.rs
```diff
@@ -15,68 +15,129 @@ use bindings::uniswap_v3_factory::UniswapV3Factory;
...
+        let slot_0 = self.inner.slot_0().call().await.unwrap();
+        self.set_pool_tick(slot_0.1);
+        self.set_pool_sqrt_price_x96(slot_0.0)
+    }
+    pub async fn _update_pool_liquidity(&mut self) {
```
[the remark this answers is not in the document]
0xJepsen: See above

crates/cli/src/main.rs
0xJepsen: Because monitor pools changes the pool by updating tick, liquidity and price.

crates/clairvoyance/src/uniswap.rs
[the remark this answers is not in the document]
ts0yu: How about we use this crate then? https://github.com/dtolnay/readonly didn't realize the reason for getter methods - sry for confusion

crates/clairvoyance/src/uniswap.rs
ts0yu: How bout we use this crate? https://github.com/dtolnay/readonly Apologies for my confusion.

crates/clairvoyance/src/uniswap.rs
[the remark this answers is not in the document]
ts0yu: should we rename it to not have the underscores and be called something like `update_pool`?

ts0yu (approved):

Issue #48, opened 2023-01-07
allow the user to have a custom config filepath


ts0yu:
we should make it so that a user can run `arbiter see --config <name>.toml` instead of only having predefined paths. it should still be able to accept `arbiter see --config`

0xJepsen:
in `arbiter/crates/cli/src/config/mod.rs` you will see this is where we parse the config file. If you run `arbiter see --config` right now it will iterate through the specified file paths of "where to look" for a file called either Config.toml or config.toml. So in the current implementation there is a specific place where the user must put their config file and it must have a specific name. thus the `--config` flag represents a boolean value that the program just checks is present. What we would like to do is all a user to "optionally" (use option type for the subcommand) specify a file name and path after the `--config` flag such that command would look something like this: arbiter see `---config somefilepathandname.toml` so that the user can specify where their config file is and what it is called. Everything for this issue is in the cli crate.

eljhfx:
Mind assigning me to this one?

Autoparallel:
@eljhfx on it.

eljhfx:
PR: #69

PR #69, opened 2023-01-14
Adding custom path for command line pool configs and Improved command line UI

eljhfx:
PR for issue #48

- Created a test config file in `crates/cli/config-test` to make sure it was working
- Changed the `struct Cli` to `struct Args to be consistent with the clap documentation
- Changed the `config` field for `enum Commands` from a `bool` to a `String` with a default value `""`
- Added logic to handle the case where a path is given as part of the `config` command such that the proper config file is used (that follows the path given in the command line)
- Reformatted the terminal output to look nicer and give users more, useful info that gives them confidence their commands were parsed correctly


crates/cli/config-test/config.toml
```diff
@@ -0,0 +1,6 @@
+rpc_url = "https://eth-mainnet.g.alchemy.com/v2/I93POQk49QE9O-NuOz7nj7sbiluW76it"
```
ts0yu, line 1: this file seems redundant as we already have an example in the config folder?

crates/cli/src/main.rs
```diff
@@ -36,9 +36,9 @@ enum Commands {
         #[arg(default_value = "5")]
         bp: String,
 
-        /// Set this flag to use the config.toml.
-        #[arg(short = 'c', long = "config", action)]
-        config: bool,
+        /// Set this flag to use a config.toml.
+        #[arg(short, long, action, default_value = "")]
```
ts0yu, line 40: it'd be nice to allow the user to just pass the `--config` flag and be able to use the default paths

crates/cli/src/config/mod.rs
```diff
@@ -53,20 +53,32 @@ pub struct Config {
...
     /// Public constructor function to instantiate a representation of a config file.
-    pub fn new() -> Result<Self, ConfigError> {
+    pub fn new(command_path: &String) -> Result<Self, ConfigError> {
         let mut content = String::new();
 
-        let config_filepaths: [&str; 2] = [
-            "./crates/cli/src/config.toml",
-            "./crates/cli/src/Config.toml",
-        ];
+        match command_path == "" {
```
ts0yu, line 59: doesn't seem to me there is any circumstance in which it is called with an empty string

ts0yu (changes requested):

ts0yu:
need to get lint passing too

crates/cli/src/config/mod.rs
```diff
@@ -53,20 +53,32 @@ pub struct Config {
 
 impl Config {
     /// Public constructor function to instantiate a representation of a config file.
-    pub fn new() -> Result<Self, ConfigError> {
+    pub fn new(command_path: &String) -> Result<Self, ConfigError> {
```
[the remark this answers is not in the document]
0xJepsen: I think it should be.

crates/cli/src/config/mod.rs
0xJepsen: Would also be taken care of it the command is an option. Still leveraging the default paths

0xJepsen (changes requested):
I think over all the implementation looks really good. I think the main thing is that we want to retain the old functionality and it the default behavior, i think making the arg an option is the right approach for this.

crates/cli/src/config/mod.rs
eljhfx: The `--config` argument is a string, the default is value is `""` (meaning if they don't input a path it's an empty string) and iterates through the default values.

crates/cli/src/main.rs
eljhfx: Ahhh will fix.

crates/cli/config-test/config.toml
eljhfx: It's a test case to check that you can set a non-default path. Can remove it if you don't think it's useful to have the test case here though.

crates/cli/src/config/mod.rs
eljhfx: Yeah was under the impression `default_value = ""` meant if you don't specify a path it won't have an error, but see this is not the case. Will fix

eljhfx:
Made the changes and ready for re-review @ts0yu

0xJepsen (approved):
looks good to me!

crates/cli/config-test/config.toml
ts0yu: oh I see, that makes sense. I don't really mind either way

ts0yu (approved):
sweet - really good stuff.

Issue #71, opened 2023-01-15
Adding error handling when Cli input is invalid


eljhfx:
When Cli inputs are spelled wrong, the code is panicking.

"thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /Users/elijahfox/Desktop/arbiter/crates/clairvoyance/src/uniswap.rs:192:37`"

We should add error handling with print statements to explain what's going on.

eljhfx:
@Autoparallel @0xJepsen 

Happy to take this one

ts0yu:
it's part of the more general error enumeration issue. we should discuss how to represent these internally

eljhfx:
What do you mean by this @ts0yu

eljhfx:
Are you suggesting it's included in #45 ? Can close it if yeah

Autoparallel:
@eljhfx It isn't mentioned directly in there. Your issue here sounds more tractable as it is smaller. The error handling should almost be a milestone as there are places to do it throughout the codebase.

eljhfx:
@Autoparallel If you want to assign this to me it probably shouldnt take more than a day or two

Autoparallel:
@eljhfx you have been assigned.

May the force be with you

PR #78, opened 2023-01-23
#71 Added Error Handling when command line input is invalid

eljhfx:
### Summary 
- In response to #71 
- Created an enum for `ClairvoyanceError` in a new file `../src/clairerrors.rs`
- Wrote logic for `Display` for the different error types (`TokenDoesNotExist, FeeTierDoesNotExist, PoolDoesNotExist`)
- Replaced `.unwrap()` with custom flow wherever possible with the appropriate errors when an error is necessary

### Future Issues
- We should probably add custom errors for other cases outside of command line inputs (also didn't get around to handling config files)


0xJepsen (approved):
this lgtm nice work @eljhfx!

PR #80, opened 2023-01-24
Alex/ci workflows

Alexangelj:
# Changes
- Adds `cargo-audit` ci workflow to run https://docs.rs/cargo-audit/latest/cargo_audit/
- Adds `cargo-deny` ci workflow to run https://github.com/marketplace/actions/cargo-deny
- Misc: fixes current CI workflow `rust` to run on `main` instead of master.

## Todo:
- There is a workflow that would make a PR for us to fix bad results from `cargo audit`, but I wasn't sure if we really wanted that so I left it out.

### DIFF
--- a/.github/workflows/cargo-audit.yaml
+++ b/.github/workflows/cargo-audit.yaml
@@ -1,11 +1,6 @@
-name: cargo-audit
+on: [push]
 
-on:
-  push:
-    branches:
-      - main
-  pull_request:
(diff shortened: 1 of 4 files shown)


Autoparallel (approved):
lgtm

Alexangelj:
> lgtm

Ay hold up - they didn't even run yet

PR #81, opened 2023-01-24
revm implementation

### LINKED ISSUE #63: implement revm for local transactions
Take the current state of some UniswapV3 pool and create a clone of this pool on a local net using Anvil.

Autoparallel:
Closes #63.

This PR made some changes to the structure of the crates. There's a new crate `simulate` and `simulation.rs` was moved into `simulate` and renamed as `price_simulation.rs`. Inside `simulate` is also `testbed` which is a start to working with `revm` and databases for simulations. 

`main()` has also changed. I added additional lines to the "Sim" command. These lines run an example that uses `testbed` and `revm` to read from a UniV2 pool.


0xJepsen:
There is a deep dependency on a package called `time` we are unable to address this until the other dependencies update their packages. 

image

0xJepsen (approved):
LGTM we just need to remember to add liscenses I made some issues for that so we can get this cooking.

PR #86, opened 2023-01-25
0xjepsen patch

### LINKED ISSUE #82: Add License to crates
This will pass the cargo-deny check we have failing right now

0xJepsen:
This closes #82 and updates our cargo-deny CI so that we pass.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,201 @@
-MIT License
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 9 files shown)


0xJepsen:
Note Cargo-audit will always fail because there is an upstream dependency on a package called `time` we can read more about the vulnerability here https://rustsec.org/advisories/RUSTSEC-2020-0071

PR #89, opened 2023-01-27
error handling

ts0yu:
problem w using panics is:
 - doesnt let library users enumerate on errors and decide how to propegate
 - defeats point of having error enum as it isnt propegated via the fn anyway

this pr
- adds more error handling
- returns custom errors for users / cli to handle


0xJepsen (approved):
This is clean i like how we propagate the contract error.

PR #102, opened 2023-02-02
Improvement/update-revm

### LINKED ISSUE #92: Update revm to 3.0.0
Revm has a new update to 3.0.0 which has breaking changes (e.g., module renames and refactors). We should not continue any further with our current version of revm so that we do not fall behind. 

This should come after #91.

Autoparallel:
Closes #92. 

Also, I took the liberty of removing unnecessary files and cargo.


0xJepsen (approved):
LGTM nice work

Issue #105, opened 2023-02-02
Deploy Contract Method


kinrezC:
- Abstract the deploy contract to revm logic into a the simulation component

Issue #90, opened 2023-01-31
simulation struct


ts0yu:
create a robust struct for all our simulation related needs:
- executing transactions
- fetching storage slots
- inserting contracts and accounts

should fit nicely w/ `architect` 

links to : https://github.com/primitivefinance/arbiter/issues/84

ts0yu:
i will work on this.

Autoparallel:
@ts0yu see what you think of the start to the `testbed` module. It will definitely see changes and I am open to modifying structures further.

ts0yu:
had a look - looks good so far. I'll work on building off of that @Autoparallel

Autoparallel:
@ts0yu, great. We'll have a much clearer idea on all of it as revm is fully implemented for sims.

ts0yu:
I would say so, yeah

0xJepsen:
I think @kinrezC is also working on some feature for the module as well

ts0yu:
what is he working on?

PR #108, opened 2023-02-03
revm stuff

ts0yu:
closes:
- https://github.com/primitivefinance/arbiter/issues/105
- https://github.com/primitivefinance/arbiter/issues/90

adds an `ExecutionManager` to abstract away functionality and make composable

should be a robust base for further simulation


Autoparallel:
@ts0yu, I think you and @kinrezC need to sync. You're both working on similar issues.

Also, I can't get the same print outputs on this branch like I do in main. Can we get those in here?

ts0yu:
oops - might of accidentally deleted something. I'll have a look tomorrow morning

Autoparallel:
Don't forget to check for lint too!

Issue #101, opened 2023-02-01
Refactor the program layout


Autoparallel:
`main.rs` should be inside of `./src/` and crates should be in their own folders in `./src/`. We should not have `main` in `./src/cli/` and instead should just call `cli` helpers from main to parse input. Ultimately `main` should be very short and easy to understand. It is too complicated right now.

We should read the [Rust book](https://doc.rust-lang.org/book/ch07-00-managing-growing-projects-with-packages-crates-and-modules.html) for guidance on this structure.

Autoparallel:
Closing as irrelevant.

PR #112, opened 2023-02-22
Moving functionality into Clairvoyance

Autoparallel:
This approaches #101 by abstracting a bit out of main.


0xJepsen (approved):
I think this is a good approach!

ts0yu (commented):
I like the general direction and am happy to merge for now but I think we need to consider the fact that clairvoyance is a library and needs to be oriented as such - maybe we could add this to utils?

Autoparallel:
> I like the general direction and am happy to merge for now but I think we need to consider the fact that clairvoyance is a library and needs to be oriented as such - maybe we could add this to utils?

This is where your input is really helpful, @ts0yu ! Please feel free to make edits and show us the best way to organize.

I was following some of your layout with Architect, after all 😄

0xJepsen (approved):
Clean love anything that reduces LOC

PR #120, opened 2023-03-09
Print help if no subcommand supplied 🐱

libevm:
Currently running `arbiter` with no subcommand provides no helpful information, this PR prints out the help log for `arbiter` when no subcommand is provided


0xJepsen (approved):
You're so smart! <3

PR #123, opened 2023-03-10
Colin/erc20 deploy

### LINKED ISSUE #84: Substitute ERC20 contracts in place of Hello World!
A better basic contract to start off with is ERC20. We will need to utilize ERC20 in nearly all simulations anyway. These contracts will have read and write capabilities. 

In this issue, let's get rid of Hello World! and deploy ERC20 as well. To close this issue, the update should involve reading from ERC20.

### LINKED ISSUE #95: Remove any use of a client/provider for offline contracts
At the moment, we are still utilizing a provider/client in generating the deployment calls. This isn't really necessary -- in no way are we actually contacting an rpc endpoint. We can get around this, but it will require getting the deployment bytecode be retrieved or constructed without appealing to `Contract::new(provider, ..)`.

Autoparallel:
Please ignore and leave the ERC20 contract in there for development purposes. 

A fix here would require generating bindings properly alongside our version of ethers. There should be a way to control versioning in forge, or just using the solc json output. Not worried about it now as this gets the contract deployed locally. At the moment we call a read function only. Next step would be testing out writing functions.

Closes #84
Closes #95 as a bonus


crates/cli/src/main.rs
```diff
@@ -92,25 +93,32 @@ async fn main() -> Result<()> {
...
-            // This is the only part of main that uses a provider/client. The client doesn't actually do anything, but it is a necessary inner for ContractDeployer
-            let contract_deployer = bindings::hello_world::HelloWorld::deploy(client, ()).unwrap();
-            let initialization_bytes = contract_deployer.deployer.tx.data().unwrap();
+            // Get a BaseContract for the ERC-20 contract from the ABI.
+            let erc20_contract = BaseContract::from(bindings::erc20::ERC20_ABI.clone());
 
-            let hello_world_contract =
-                BaseContract::from(bindings::hello_world::HELLOWORLD_ABI.clone());
+            // Choose name and symbol for the constructor args required by ERC-20 contracts.
+            let name = "ArbiterToken";
```
kinrezC, line 103: We can address this in a later PR, but if it's going to be too challenging to do the general deployer abstraction we can at least do it for `deploy_token` or something.

kinrezC (approved):
LGTM

crates/cli/src/main.rs
Autoparallel: We can certainly do a general deploy abstraction! I'd enjoy making that actually.

0xJepsen (approved):
amazing work, very nice.

PR #124, opened 2023-03-13
Colin/erc20 mint

### LINKED ISSUE #94: Write using ERC20
We need to be able to write to contracts. Given that (#84) is closed, we need to create a PR that writes to the ERC20 contract using a eth call and updates the state.

Autoparallel:
First off,  lots of changes here, sorry. 

I was having issues getting input tokens to work properly and figured out that error by writing the test in `simulate/lib.rs`.  This involves the new contract `writer.sol`. This could be extended to check that all input types we can send to functions will be valid. Some of this could just be copied from what's in `main.rs` now as well.

I then wanted some ability to mint ERC20 token, so I created an instance of the contract called `arbiter_token.sol` which has a very trivial implementation of `mint` that calls the internal `_mint` function in `ERC20.sol`. At this point, I just decided to include the contracts I've been using in the directory `contracts/`. I generated very minimal bindings for some new contracts due to Ethers version issues and Foundry issues. It all works, but that is janky and we should clean this up at some point.

This closes #94.


crates/simulate/src/lib.rs
```diff
@@ -1,2 +1,74 @@
...
+
+        // Deploy the writer contract.
+        manager.execute(
+            user_address,
+            writer_bytecode,
+            TransactTo::create(),
+            Uint::from(0),
+        );
+
+        let writer_contract_address = manager
```
0xJepsen, line 32: So functional

crates/simulate/src/lib.rs
```diff
@@ -1,2 +1,74 @@
...
+            .accounts
+            .into_iter()
+            .nth(2)
+            .unwrap()
+            .0;
+
+        // Generate calldata for the 'echoString' function
+        let test_string = "Hello, world!";
+        let input_arguments = test_string.to_string();
+        let call_bytes = writer_contract.encode("echoString", input_arguments);
```
0xJepsen, line 46: this is genius

0xJepsen (approved):
Nice work on this.

PR #125, opened 2023-03-14
Colin/deploy contract abstraction

### LINKED ISSUE #105: Deploy Contract Method

Autoparallel:
Took a stab at abstracting the deployment process. This could surely be cleaned up more, but it gets the job done!

I created a struct `SimulationContract<Deployed>` to mimic (in some way) the ethers contract struct but without any provider. You can begin by calling `SimulationContract::new(...)` which will return `SimulationContract<NotDeployed>`.  Only after calling `ExecutionManager::deploy(...)` will you receive back a `SimulationContract<IsDeployed>`. The contract will also be in the relevant db.

Closes #105

### DIFF
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1,4 +1,3 @@
-use std::{env, str::FromStr, sync::Arc};
 use clairvoyance::Clairvoyance;
 use clap::{CommandFactory, Parser, Subcommand};
 use ethers::{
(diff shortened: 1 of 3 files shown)


0xJepsen (approved):
This looks like a good start. Maybe we can jam together on how we should clean things up more.

PR #127, opened 2023-03-15
Waylon/portfolio deploy

### LINKED ISSUE #96: Deploy the UniV3 factory contract
Now that we have ERC20 up, we can deploy the UniV3 factory contract. To close this issue, let's also require that we initialize a pool with two ERC20 tokens.

### LINKED ISSUE #128: Remove Clairvoyance
This feature is tech dept and we need to pull it out.

0xJepsen:
While working on this i have some ideas on how we can maybe depreciate some older features (tech dept) and also do better contract management so that our bindings are always generated with the same versions. 

I did also recently discover that the new version of ethers (2.0.0) has been fixed and actually works nicely. However there are breaking changes so that the old binds and new bindings do not play nicely together (second suggestion). There is one small error in the new bindings https://github.com/gakonst/ethers-rs/pull/2263 but it doesn't effect us and looks like there is already a pr up for a fix. 

My suggestions
- [x] Pull out clairvoyance. Reason for this is that it's not modular and doesn't interact with revm so isn't really our priority right now. Also it is currently broken with the new bindings. 
- [ ] Keep a directory in the root of the project for all contracts we want bindings for. This will help us ensure they are all always generated with the same version of abigen
- [ ] Deploy Portfolio to revm with these new bindings 

Closes #96
Closes #128


Alexangelj:
This makes sense to me - really sucks the ethers dependencies / bindings are so messy

0xJepsen:
> This makes sense to me - really sucks the ethers dependencies / bindings are so messy

Yeah they usually aren't i think part of it was our fault not keeping the contracts here which created some fragmentation on the versioning. Then there was a breaking change in the new ethers version which is addressed now. 

I think in general the bindings are incredibly helpful

PR #130, opened 2023-03-16
simplified the deploy function

Autoparallel:
The execution result contains the newly generated address. Let's just use this.


0xJepsen (approved):
This is much cleaner. Nice work

PR #132, opened 2023-03-16
Waylon/bindings

0xJepsen:
This sorts out all bindings. From now on we can regenerate up to date bindings from the contracts located in the `src/` directory. I did have to migrate two old bindings to `crates/bindings_old/` so that clairvoyance was happy.


Autoparallel (approved):
lgtm

PR #134, opened 2023-03-17
Waylon/gitmodules

### LINKED ISSUE #96: Deploy the UniV3 factory contract

### LINKED ISSUE #133: Migrate testing contracts to their own repo
We should be using contracts as git submodules, rather than locally.

0xJepsen:
Closes #133 
Closes #96


0xJepsen:
Portfolio should have the correct constructor args but i hit this invalid type when tokenizing the args. I am wondering if i need to figure out the right rust type to have them tokenized as an `Address`. 

Screenshot 2023-03-16 at 9 13 39 PM

PR #141, opened 2023-03-20
edits that made environment struct more private

Autoparallel:
Just a refactor. Should prevent some foot bullets for people in the future


0xJepsen (approved):
This is clean

PR #151, opened 2023-03-21
Colin/result handler

### LINKED ISSUE #148: Result handler for the execution logic in `execution.rs`
We want to abstract away redundant logic so we don't have to repeat and make unnecessary LOC. This should be implemented by the simulation environment. 

- [ ] This function should match the output and do appropriate error handling on result types. 
- [ ] If the result type is an event object we should call the abstraction logic in #143

Autoparallel:
Closes #148.

There's some other work I put in here but remains to be implemented. It will be easier to get everything working together properly as we continue to work.


0xJepsen (approved):
very based

PR #152, opened 2023-03-22
Implemented an event buffer

Autoparallel:
Event logging will be important for agents to act upon. So, what I tried to do here was implement a buffer called `event_buffer` that the `SimulationEnvironment` owns (as an attribute) write access to and, upon execution, echoes out logs into this buffer. Other agents can freely read at any time. Due to this write protection, the buffer is not mutex. We can get away with `Arc<RwLock<...>>`.

I'm not sure this is the best way to handle it, but it seems to work well for now! Some other changes came along with this. 

To see this in action, just investigate `test_event_logging` in Simulation's lib.

### DIFF
--- a/crates/simulate/src/environment.rs
+++ b/crates/simulate/src/environment.rs
@@ -18,7 +18,7 @@ use crate::agent::{Agent, TransactSettings};
 struct SimulationEnvironment {
     evm: EVM<CacheDB<EmptyDB>>,
     _agents: HashMap<String, Box<dyn Agent>>,
-    event_buffer: Arc<RwLock<Vec<Vec<Log>>>>,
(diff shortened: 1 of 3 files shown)


crates/simulate/src/environment.rs
```diff
@@ -58,8 +58,10 @@ impl SimulationEnvironment {
         if let Some(handle) = self.writer_thread.take() {
             handle.join().unwrap();
         }
-        println!("Writing logs: {:#?}", logs);
-        self.event_buffer.write().unwrap().push(logs);
+        self.event_buffer.write().unwrap().clear();
+        logs.into_iter().for_each(|log| {
```
0xJepsen, line 66: Nice Closure here

0xJepsen (approved):
This looks good to me!

PR #160, opened 2023-03-23
Colin/bindings cleanup

### LINKED ISSUE #149: Make bindings default output to our crate
In the `foundry.toml` We want to make the default output fit our needs and generate the bindings in the bindings crate with the `cargo.toml` that is desired and package names we are working with.

Autoparallel:
Diffs are large, but it is because we are just adding a lot of contracts. We don't need them all, but they are far easier to regenerate now with just two commands. We can work on making this even easier in the future. The commands are described now in the updates to README. Also, main was adjusted slightly for renames and to include bindings along with other crates/modules.

Closes #149


0xJepsen (changes requested):
I think we need the license in the bindings crate for deny to pass.

0xJepsen (approved):
amazing clean

PR #165, opened 2023-03-26
Fixed GBM in price_simulation.rs

### LINKED ISSUE #163: Fix the price simulation
Currently this feature is broken. When running `arbiter gbm` we get errors. We need to fix this so we can see the output of the GBM price path in some figure. Perhaps we also write a test here.

0xEstelle:
fixed the GBM process in price_simulation.rs and changed the config to match units of time between sigma timestep and mu. Closes #163

### DIFF
--- a/crates/cli/src/config.toml
+++ b/crates/cli/src/config.toml
@@ -6,11 +6,11 @@ token1 = "USDC"
 bp = "30"
 
 [sim]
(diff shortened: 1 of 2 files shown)


0xJepsen (approved):
Estelle this is great! Nice work!

PR #176, opened 2023-03-28
refactoring manager and environment

ts0yu:
clean some stuff up

PR #179, opened 2023-03-29
issues #176 and #177

### LINKED ISSUE #177: Add an `admin` agent to assume some of the roles currently given to `SimulationManager`
Right now this is a bit broken. Agents need to be able to make calls to the EVM just like the manager can. 

- [ ] Create an `admin` agent that assumes some of the current roles of the `SimulationManager` (This could make the most sense if the `SimulationManager` should just run sims and not be involved with the sims themselves.)
- [ ] Define a generic `call` function for agents that are in a `SimulationEnvironment`.

### LINKED ISSUE #178: Create a basic `User` agent
The `User` should be used as a dummy agent that can make calls, but does not do so procedurally.

- [ ] Create a `User` struct with an address and transaction settings.
- [ ] Implement `Agent` for this user.
- [ ] Reorganize the agent capabilities as needed (see #177 as well).

Autoparallel:
Closes #177, closes #178

This PR closes the above issues by letting the whole `SimulationEnvironment` be under an `Arc<RwLock<SimulationEnvironment>>`. We also now place a reference to this environment inside of each agent that is given access to write (single thread access protected) and read (multithread access). I created the `Admin` agent and a `User` agent which are very similar. `Admin` spawns along with a `SimulationEnvironment` so it can be used as a default deployer and `Admin` will always assume address 1,  is named `"admin"`, and can be given as a reference by calling `manager.admin()`. The `User` can be created with any `name` and `Address` (`B160`).  We use the `User` in the `test_token_mint` test.

~~I am getting stuck with lifetimes and reference counters. There are some things to do still.~~

~~What my goals are is to give every single agent access to a mutable copy of the `SimulationEnvironment` that a `SimulationManager` controls. This is because the environment is something all agents will need to access and make transactions with. No one agent should hold on to the environment, but they all need to share a reference to the same one.~~

~~There's some other reorganization I've done. I have tried a few things that also attempted just using references and specified lifetimes but this is the most successful so far.~~


Autoparallel:
Might be able to avoid using trait objects (dyn Agent) inside of the agents hashmap. Could just use a generic here?

Autoparallel:
I'm trying out a RwLock on SimulationEnvironment in the same way we do logs as this works well and follows the same design principal of a "permissioned writer" and "permissionless readers". Right now the test `test_token_mint` is failing and I am trying to see why.

PR #181, opened 2023-04-04
Colin/agents-threaded-event-reader

### LINKED ISSUE #153: Multithreaded event handling for agents
This issue came up in thinking about how to allow agents to asynchronously await events in order to dictate their actions. There are likely multiple ways to accomplish this, but here's an outline.

- [ ] Allow the `SimulationEnvironment` to own write control over a buffer where events are written. Upon any call to execution, events should be echoed into this buffer.
- [ ] Agents need read only access to the buffer and as such they can read from this buffer without the buffer needing to be mutex since agents will not write -- only the environment/EVM instance will write. Agents need a function to access these logs.
- [ ] Agents will also need some type of async await function that lets them constantly poll for updates in event logs.

Autoparallel:
Attempting to close #153. 

Still need to build the functionality of actually awaiting new events. Trying to model this after ethers.

### DIFF
--- a/crates/simulate/src/lib.rs
+++ b/crates/simulate/src/lib.rs
@@ -9,7 +9,7 @@ pub mod utils;
 
 #[cfg(test)]
 mod tests {
-    // use core::slice::SlicePattern;
+    use tokio::test;
     use std::str::FromStr;
 
     // use bindings::{self, arbiter_token};
@@ -26,7 +26,7 @@ mod tests {
     #[test]
     /// Test that the writer contract can echo a string.
     /// The writer contract takes in no args.
-    fn test_string_write() {
+    async fn test_string_write() {
         // Set up the execution manager and a user address.
         let mut manager = SimulationManager::default();
         let admin = manager.admin();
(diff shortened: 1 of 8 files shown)


0xJepsen (commented):
This looks good to me other than the lint failing, what other things need to be done on this PR?

Autoparallel:
> This looks good to me other than the lint failing, what other things need to be done on this PR?

It was not very clean at that point -- just working. I cleaned it up and it's looking good now. Opening PR now.

Autoparallel:
Closes #153

Autoparallel:
I need to set up dispatching so messages aren't consumed on other threads. Working on this now.

Autoparallel:
I think this should close it. See the reworked test.

crates/simulate/src/lib.rs
```diff
@@ -316,6 +316,57 @@ mod tests {
             }
         });
 
+        let reader_for_admin = manager.agents.get("admin").unwrap().receiver();
+        let writer_base_contract_for_admin = writer.base_contract.clone();
+        let admin_handle = thread::spawn(move || {
+            let mut i = 0;
+            while let Ok(logs) = reader_for_admin.recv() {
```
0xJepsen, line 323: nice

0xJepsen (approved):
lgtm!

PR #195, opened 2023-04-11
Colin/liquid exchange

### LINKED ISSUE #154: Infinitely liquid market
In order to perform arbitrage, we will need at least two exchanges. The most basic exchange is *infinitely liquid*, e.g., it assumes no price impact. We can implement this exchange under `manager` control.

- [ ] Write a very basic contract for this exchange and give access to the `setPrice` (or similar) function (for some token pair) only to the contract creator (in our case, this is always the `manager`). 
- [ ] Contract should echo the new price as an event.
- [ ] Contract should allow for swapping a token at whatever price it is listed at for the other token in the pair. This will require minting on behalf of the `manager` as well. 

Prior to completion of this contract, it will be important to set up an exchange trait in order to keep the way we interact with exchanges generic.

Autoparallel:
The liquid exchange is alive now. Two tests check that the swap `x->y` and `y->x` function properly. Outlined an `exchange.rs` module that we can bind to later on.

Closes #154


crates/simulate/src/exchange.rs
```diff
@@ -0,0 +1,335 @@
+#![warn(missing_docs)]
+//! This module contains the `Exchange` and `Cfmm` traits that are used to describe the functionality of a contract that can be used to swap tokens.
+
+use ethers::types::Address;
+
+/// A trait that describes the functionality of any exchange.
+pub trait Exchange {
```
0xJepsen, line 7: Yee, and i cannot express this enough, Haw

0xJepsen (approved):
Very nice work

PR #197, opened 2023-04-12
removed admin submodule

### LINKED ISSUE #183: Remove `admin` submodule in favor of `user`
We can remove the `admin` submodule and just use the `user` submodule to reduce boilerplate.

- [x] When we create a new `SimulationEnvironment`, just call `User::new()` with the default admin credentials.
- [x] Make sure that when we `create_user()` that we report errors on naming and address collisions.

Autoparallel:
This closes #183.


0xJepsen (approved):
Smells good in here

PR #202, opened 2023-04-14
Improving upon contract interfaces

Autoparallel:
- [x] Created a minimal implementation of `SimulationMiddleware` that will let us use the forge generated bindings for a nicer API.
- [x] Moved the `SimulationContract` struct and related items into the `contract.rs` module and out of `environment.rs`.
- [x] Got rid of the option for the address in the `SimulationContract` using types.


Autoparallel:
Clippy won't pass until we clean up the simulation price generator stuff.

0xJepsen (approved):
Delicious!

PR #206, opened 2023-04-14
Historical Data

### LINKED ISSUE #201: Back Testing Data
Get price path from Ethereum mainnet for backtesting. This will be
- [ ] Swap data for stable pools from uniswapV3 over 1 day

0xEstelle:
Closes #201 . This is not really done yet, all I have as of the init comment of this PR is a new impl for filtering past data instead of live monitoring. Going to write some tests and integrations next.


Autoparallel (approved):
lgtm

PR #207, opened 2023-04-15
Colin/cleanup

Autoparallel:
Lots of various clean up in here to reduce LoC and make some interactions more simple than they were before. Biggest upgrades are to the `SimulationContract` which now has more differences between the states of `NotDeployed` and `IsDeployed`. Some improvements to error handling with the `SimulationManager`. Also, I introduced the `AgentType` enum which gets rid of the need for dynamically dispatched agents which should be a bit nicer.

### DIFF
--- a/crates/simulate/src/contract.rs
+++ b/crates/simulate/src/contract.rs
@@ -4,8 +4,8 @@ use std::marker::PhantomData;
 
 use bytes::Bytes;
 use ethers::{
-    abi::{Constructor, Contract, Detokenize, Token, Tokenize},
-    prelude::{AbiError, BaseContract, Lazy},
+    abi::{Contract, Detokenize, Token, Tokenize},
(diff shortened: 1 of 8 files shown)


crates/simulate/src/contract.rs
```diff
@@ -17,49 +22,107 @@ pub struct IsDeployed;
...
             deployed: std::marker::PhantomData,
+            base_contract: (),
+            constructor_arguments,
         }
     }
 }
+
+impl SimulationContract<IsDeployed> {
+    /// Encodes the arguments for a function call for the [`SimulationContract`].
+    pub fn encode_function(&self, function_name: &str, args: impl Tokenize) -> Bytes {
```
0xJepsen, line 102: This is nice

0xJepsen (approved):
this is good QoL improvements

PR #217, opened 2023-04-19
Stochastic.rs Restructure

### LINKED ISSUE #209: Refactor `price_simulation.rs`
This module has improved but we can take it further and get a good first issue from it. I also am proposing that the name should be changed. Elsewhere we have things like `SimulationManager`, `SimulationEnvironment`, and `SimulationContract`. Perhaps we can go with `SimulationPricePaths`? In which case, maybe we just go with `price_path.rs`? First task:

- [x] Change the name to something better! (We will approve this during the PR process.)

We may end up with a few different price path generators but each generator will have a lot of overlapping properties. Each of which should input a struct that captures these properties. This will decrease the input size to create each new price path.

- [x] Create a struct whose fields capture attributes that **_all_** price paths will have (e.g., timestep, num_steps, initial_price, seed, etc.).
- [x] (Potentially) create a `PriceProcess` trait that gives the basic functions that each price path needs to implement. This could be overkill if there is just a single function (e.g., generate). Not overkill if we have many processes, though.
- [x] Create structs for each type of `PriceProcess`. These structs should be tailored to the *specific* properties that the `PriceProcess` has. For example, drift/volatility for GBM and mean/reversion_speed for OU.
- [x] Create an enum for these price paths. (The `PriceProcess` trait can be implemented on this enum if we go this route.)

These changes will likely cause breaks in other areas (such as the `test_price_simulation_oracle` in `exchange.rs`).

0xEstelle:
Closes #209. Not really close to done yet just started the restructure. As of this point I've made a `Distribution` trait with one `sample` method along with a series of other distribution structs that implement the trait (everything from `normal` to `gamma` and `beta`). Also have initialized the process of doing the same thing with different types of price paths and a more generic trait `PriceProcess`. Will update progress here

### DIFF
--- a/crates/simulate/src/stochastic/mod.rs
+++ b/crates/simulate/src/stochastic/mod.rs
@@ -1,6 +1,8 @@
 use rand::Rng;
 use rand::distributions::{*};
 
+pub mod price_process;
+
 /// A trait for distribution sampling.
 pub trait Distribution<T> {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> T;
--- a/crates/simulate/src/stochastic/mod.rs
+++ b/crates/simulate/src/stochastic/mod.rs
@@ -4,11 +4,8 @@
 
 use rand::Rng;
 use rand_distr::{
-    Normal as NormalDistr,
+    Beta as BetaDistr, Exp as ExpDistr, Gamma as GammaDistr, Normal as NormalDistr,
     Poisson as PoissonDistr,
-    Exp as ExpDistr,
-    Gamma as GammaDistr,
-    Beta as BetaDistr,
 };
 
 pub mod price_process;
--- a/crates/simulate/src/stochastic/mod.rs
+++ b/crates/simulate/src/stochastic/mod.rs
@@ -1,10 +1,21 @@
+//! Distribution sampling and price process generation and plotting.
+
+#![allow(dead_code)]
+
 use rand::Rng;
-use rand::distributions::{*};
+use rand_distr::{
+    Normal as NormalDistr,
+    Poisson as PoissonDistr,
+    Exp as ExpDistr,
+    Gamma as GammaDistr,
+    Beta as BetaDistr,
+};
 
 pub mod price_process;
 
 /// A trait for distribution sampling.
 pub trait Distribution<T> {
+    /// Sample from a distribution
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> T;
 }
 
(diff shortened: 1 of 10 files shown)


crates/simulate/src/stochastic/price_process.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Generates a GBM price path.
+    /// # Arguments
+    /// * `timestep` - Time step of the simulation. (f64)
+    /// * `num_steps` - Number of steps in the simulation. (usize)
+    /// * `initial_price` - Initial price of the simulation. (f64)
+    /// * `seed` - Seed for testing. (u64)
+    /// # Returns
+    /// * `time` - Vector of time steps. (Vec<f64>)
+    /// * `prices` - Vector of prices. (Vec<f64>)
+    pub fn generateGBM(&self, timestep: f64, num_steps: usize, initial_price: f64, seed: u64) -> (Vec<f64>, Vec<f64>) {
```
Autoparallel, line 104: Are you planning to just pass a `Price` struct into this? That could be a good way to cut down on the number of inputs.

crates/simulate/src/stochastic/price_process.rs
```diff
@@ -0,0 +1,184 @@
...
+        (time, prices)
+    }
+}
+
+/// Import CSV file of price data
+/// # Arguments
+/// * `file_path` - path to csv file of price data (String)
+/// # Returns
+/// * `price_data` - Vector of prices. (Vec<f64>)
+pub fn import_price_from_csv(file_path: &str) -> Result<Vec<f64>, Box<dyn Error>> {
```
Autoparallel, line 173: Nice call on this.

crates/simulate/src/stochastic/price_process.rs
0xEstelle: yes I am! currently working on it along with integration, will push recent changes today

0xEstelle:
all working, ready for yalls reviews

crates/simulate/src/stochastic/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Create a new poisson distribution
+    fn new(mean: f64) -> Poisson {
+        Poisson { mean }
+    }
+}
+
+impl Distribution<f64> for Poisson {
+    /// Sample from a poisson distribution
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> f64 {
+        let poisson = PoissonDistr::new(self.mean).unwrap();
```
Autoparallel, line 55: This might not be a good idea. You're going to make a new distribution each time you want to sample? I think we could just have a distribution type enum so that we can select distributions and also avoid recreating a distribution each time we want to sample. What do you think?

Autoparallel: Also if you already have a distribution created in `new()` for `impl Poisson` (for example), then you can just do `self.sample()` basically

Autoparallel (commented):
Left some comments

Autoparallel (approved):
lgtm, I think in the future I'd not abbreviate any names but that's my only nit pick!

PR #224, opened 2023-04-25
Waylon/versioning

### LINKED ISSUE #223: Automatic Versioning
Look into ways to continually push minor releases automatically. Could be a nice solution with CI workflows.

0xJepsen:
Closes #223

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   bump_version:
     runs-on: ubuntu-latest
-    if: github.event.pull_request.merged == true && (github.event.number % 5 == 0)
+    if: github.event.pull_request.merged == true
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  bump_version:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.merged == true && (github.event.number % 5 == 0)
```
Autoparallel, line 11: Honestly we can probably just bump the patch version every single PR instead of every 5. We are only a 100 or so PRs in and I imagine we'll only get 10-50 patch versions per minor version.

.github/workflows/bump_version.yml
```diff
@@ -0,0 +1,43 @@
+name: Bump minor version
```
Autoparallel, line 1: This again should bump the patch version not the minor version

Autoparallel (changes requested):
I'm a bit nervous to merge this at the moment as I'm not able to parse the logic completely. Where are we actually incrementing the patch version by 1? I know it's in lines 37-42, but I don't get it.

ankit-gautam23 (approved):
LGTM

Autoparallel (approved):
lgtm

PR #230, opened 2023-04-27
Colin/simple arb detection

### LINKED ISSUE #144: Agent Type: SimpleArbitrageur
In the `/arbiter/crates/simulate/src/` directory, implement an `SimpleArbitrageur` module/struct.
- [x] Consume and monitor events #142 

`SimpleArbitrageur`:

- [x] The `SimpleArbitrageur` struct has filters and stores prices.
- [x] Spawn a thread that listens for events that match the filters.
- [x] Compute the price differences between each pool after each price change and store them.
- [x] When the prices are detected to be different, exit the thread.

Autoparallel:
Closes #144 

Quite a big refactor was needed in order to give us the ability to have a struct like `SimpleArbitrageur` an impl block that we can access from a simulation. That is, working with `&dyn Agent` can't do this and it would have required a complicated `event_monitoring()` function to be created. 

This is now done by using the `AgentType` enum which does create an onion type by virtue of items like `AgentType::User(User)`. This is not the most elegant, but we get more control this way.

In order to make the creation of agents easier, we have a trait called `AgentStatus` that allows for fields of each agent struct to change depending on whether it has been added to the `SimulationManager`'s HashMap as well as the `SimulationEnvironment` db. This also required a bit of finagling  with another trait `Identifiable` which solely retrieves the name of an agent.

A new test was written in `simple_arbitrageur.rs` called `simple_arbitrage_detection()`. Think of this as a basic boilerplate example of detecting price differences. It is rather clunky for now, but going further felt a bit silly until we have other pools to arb with.


Autoparallel:
Just fixed the conflicts.

crates/simulate/src/agent/mod.rs
```diff
@@ -131,14 +197,29 @@ pub trait Agent: Sync {
...
+    // TODO: add a condition as a bool valued function?
+    // TODO: It would be optimal to not build functions inside of the monitor events since it could get called often. Ideally we just don't call it often?
     /// Monitor events for the agent.
     fn monitor_events(&self) {
         let receiver = self.receiver();
+        let event_filters = self.event_filters();
         thread::spawn(move || {
-            while let Ok(_logs) = receiver.recv() {
-                // self.filter_events(logs);
+            let decoder = |input, filter_num: usize| {
```
0xJepsen, line 208: sick move

crates/simulate/src/agent/mod.rs
```diff
@@ -151,23 +232,52 @@ pub struct SimulationEventFilter {
     pub address: B160,
     /// The event names to filter for.
     pub topic: H256,
+    /// A private copy of the [`BaseContract`] for whichever contract is used to generate filters.
+    base_contract: BaseContract,
```
0xJepsen, line 236: why does this need to be private?

0xJepsen (approved):
Looks good to me!

PR #238, opened 2023-04-30
Fixes issue #234

### LINKED ISSUE #234: Make an install script
Our README has a slight error -- `cargo build` will generate the debug version of the `arbiter` binary, but it will not add this to the user's path. We should make an install script that works for UNIX and Windows environments. Here's a potential solution I haven't tried that I would like a community member to clean up and submit as a PR.

```bash
#!/bin/sh

# Build release binary
cargo build --release

# Install binary to system
case "$(uname -s)" in
    Linux|Darwin)
        # Unix-based systems (Linux, macOS)
        TARGET_DIR="/usr/local/bin"
        echo "Installing to $TARGET_DIR..."
        sudo ln -sf "$(pwd)/target/release/arbiter" "$TARGET_DIR/arbiter"
        ;;

    MINGW*|CYGWIN*|MSYS*)
        # Windows with Git Bash or WSL
        TARGET_DIR="/c/Program Files"
        echo "Installing to $TARGET_DIR..."
        cp "$(pwd)/target/release/arbiter.exe" "$TARGET_DIR/arbiter.exe"
        ;;

    *)
        # Unsupported system
        echo "Unsupported system. Please install the binary manually."
        exit 1
        ;;
esac

echo "Installation successful. You can now use 'arbiter <command>'."
```

To close this issue, do the following:

- [ ] Make a script called `install.sh` that will build the release optimized `arbiter` binary.
- [ ] Have the script add the `arbiter` binary to the user's PATH regardless of UNIX/Windows environment.
- [ ] **Verify this works at least on MacOS, Windows, and Ubuntu.**
- [ ] Make the necessary changes in the README to guide a user through this installation process.

Bonus points if you add ASCII art or make this process even simpler!

MakisChristou:
I have added a slight modification to the `install.sh` to use the rust bin directory that is already created and added to the path automatically on Windows.

Closes #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,20 +35,25 @@ For our next beta release, we will be focusing on the following features:
 
 First, clone the repository to your local environment so
 
-```
+```bash
 git clone https://github.com/primitivefinance/arbiter.git
 cd arbiter
-cargo build
+```
+
+Install arbiter on your system:
+
+```bash
+cargo install --path crates/cli --force
 ```
 
 With the `arbiter` binary generated, you can run commands such as:
-```
+```bash
 arbiter sim
 ```
 
 ## Generating Docs
 To see the documentation for Arbiter, after cloning the repo, you can run:
-```
+```bash
 cargo doc --workspace --no-deps --open
 ```
 This will generate and open the docs in your browser. From there, you can look at the documentation for each crate in the Arbiter workspace.
@@ -57,7 +62,7 @@ This will generate and open the docs in your browser. From there, you can look a
 
 In the `lib/` folder you can add additional smart contracts or regenerate Rust bindings. To do so, run the following from the Arbiter root directory:
 
-```
+```bash
 ./bind.sh
 ```
 


Autoparallel:
I don't have a Windows machine to test this on personally. @0xEstelle or @Alexangelj can you give this a shot?

Autoparallel:
> I have added a slight modification to the `install.sh` to use the rust bin directory that is already created and added to the path automatically on Windows.

Hi @MakisChristou, I'm curious what you think about a slightly different approach. With cargo, we can instead do:

```bash
cargo install --path crates/cli --force
```

This will install the binary in 
`Users/$USER/.cargo/bin/arbiter`

From my experience, this also generates a symlink to `arbiter`. 

This can still be put in an `.install.sh` script to make life a bit easier. 

What do you think?

MakisChristou:
Hello @Autoparallel. I have tried this on both MacOs and Windows 11 and it looks like its working. So maybe we can just skip the `install.sh` script altogether? I don't see why its needed.

Autoparallel:
> Hello @Autoparallel. I have tried this (`cargo install --path crates/cli --force`) on both MacOs and Windows 11 and it looks like its working. So maybe we can just skip the `install.sh` script altogether? I don't see why its needed.

Seems reasonable. Perhaps a good way to give you credit through this work is to edit the README with these instructions?

MakisChristou:
What about now?

Autoparallel:
Perfect. I'll merge now.

Autoparallel (approved):
lgtm

PR #245, opened 2023-05-02
Waylon/sim mod

### LINKED ISSUE #239: Modular Sims
I think one thing we should do in our architecture is make a a simulation crate where we can have on going constructions on a collection of protocols like `portfolio`, `uniswap`, `curve`, etc. These will 
- Serve as an example of the power of arbiter to external contributors and projects
- Demonstrate economic risk auditing capabilities
- Give more opportunities from other core protocols to contribute

I think the best way in which we do this needs more discussion. My initial idea is that we change `sim.rs` to `portfolio_sim.rs` and then put it in a sims module where we have sims for major core protocols. I think the other sims will actually be much easier to build and design as well since portfolio has custom bit encoding for it's calls which introduce a lot of developer frictions and overhead. My open questions are:
- where in the code base should this module live? Right now the `sim.rs` file lives in `crates/cli/src/commands` maybe we make a `crates/cli/src/sims` directory for this to live in? What are the teams thoughts.

### LINKED ISSUE #240: Create `sim` as a module with submodules.
Based of #239 we need additional modularity in creating new simulations as we move forward. A step in the right direction is to create a `sim` (sub)module.

I propose the following but am willing to discuss alternatives:

- [x] Replace `sim.rs` in `crates/cli/src/commands` with `crates/cli/src/commands/sim/mod.rs`.
- [x] Take the current code in `sim.rs` which is specific to an *in progress* build of a simulation for Portfolio contracts and place this code in `crates/cli/src/commands/sim/portfolio.rs`

The `mod.rs` file should be structured as a module.

0xJepsen:
Closes #240 

This migrates the sims to there own module and starts an example skeleton for uniswap sims (which imo makes sense since we also have their bindings already). @Autoparallel I think this really opens up a nice landscape for some more issues here, also gives incentive for devs from other teams to get their hands dirty with arbiter and utilize it.


0xEstelle (approved):
Looks good to me! this is exciting

Autoparallel (approved):
Lgtm. 

I think I have an issue in mind that would be a good first issue for a new comer to tackle after I merge this.

PR #248, opened 2023-05-03
Waylon/new portfolio abi

### LINKED ISSUE #225: Allocate and Deallocate encodings and calls
now that we have a pool from #216 we need to validate the correct encoding for the the allocate and deallocate calls. This involves
- [x] Write test
- [x] Get tests to pass (need to figure out the right parameters to give)

### LINKED ISSUE #247: New Portfolio Abi
We should rewrite our calls to portfolio with the new abi which will 
1) make our lives much easier 
2) be the final spec for our contracts

Will involve 
- [x] installing the right [submodule branch](https://github.com/primitivefinance/portfolio/commit/f6b761a81c5207ac9cc2d24ee80b2b83ad5d2b51) from portfolio 
- [x] regenerating bindings
- [x] taking out all the encoding logic in the sims
- [x] replacing the calls with the new abi

0xJepsen:
Closes #247 
Would like to also Close #225, but need to chat a little bit about the right process to get the delta_liquidity parameter
@kinrezC Would you like to chat about this after standup?


Autoparallel:
Careful also that this is currently closing #225

0xEstelle (approved):
this is so clean yay!

Autoparallel:
@0xJepsen @0xEstelle the last commit here 77854f2 deletes `LiquidExchange.sol`.

PR #249, opened 2023-05-03
Estelle/objectives

### LINKED ISSUE #242: Implement the `LinearNonnegative` objective function
This would populate its own submodule titled something along the lines of `linearnonnegative.rs` contained in the module `objective/mod.rs` contained within the `on-chain` crate. This can broken down into two tasks, 
- [x] Implement the struct `LinearNonnegative` to initialize the variables at play
- [x] Implement the conjugate of the utility function of `LinearNonnegative`, call it something like `fn`
- [x] Implement the gradient function of `fn`
- [x] Implement value limits on Lagrange multiplier `v` associated with the objective

Can use both [CFMMRouter.py](https://github.com/primitivefinance/CFMMRouter.py) and [CFMMRouter.jl](https://github.com/bcc-research/CFMMRouter.jl) as a reference (they each have objective source files). Each function will contain an infinity instance, which needs more information on how to approach in Rust.

### LINKED ISSUE #243: Implement the `BasketLiquidations` objective function
this would populate its own submodule titled something along the lines of `basket_liquidation.rs` contained inside the module `objective/mod.rs` contained in the `on-chain` crate. This can be broken down into two tasks:
- [x] Implement the struct `BasketLiquidations` initializing the variables at play
- [x] Implement the conjugate of the utility function of `BasketLiquidations`, call it `fn`
- [x] Implement the gradient function of `fn`
- [x] Implement the value limits of the Lagrange multiplier `v`

Can use both [CFMMRouter.py](https://github.com/primitivefinance/CFMMRouter.py) and [CFMMRouter.jl](https://github.com/bcc-research/CFMMRouter.jl) as a reference (they each have objective source files). Each function will contain an infinity instance, which needs more information on how to approach in Rust.

0xEstelle:
Closes #242 and Closes #243. Not entirely sure which Type we expect to get pool price data so i implemented everything as f64 and Vec<f64>. Can change if we need!

### DIFF
--- a/crates/on-chain/src/objectives/basket_liquidation.rs
+++ b/crates/on-chain/src/objectives/basket_liquidation.rs
@@ -1,0 +1,62 @@
+use std::f64::{INFINITY, EPSILON, sqrt};
+
+/// Basket Liquidation objective function.
+pub struct BasketLiquidation {
+    /// Index token to swap into.
+    i: usize,
+    /// DeltaIn vector of assets.
+    delta_in: Vec<f64>,
+}
+
+impl BasketLiquidation {
+    pub fn new(i: usize, delta_in: Vec<f64>) -> Self {
+        Self { i, delta_in }
+    }
+    /// Conjugate utility of the Basket Liquidation objective function.
+    pub fn conjugate_utility(&self, v: Vec<f64>) -> f64 {
+        if v[self.i-1] <= 1.0 {
+            let mut vec = Vec::new();
+            for i in 0..self.delta_in.len() {
+                if i == self.i-1 {
+                    vec.push(0.0);
+                    } else {
+                    vec.push(self.delta_in[i]*self.v[i]);
+                }
+            }
+            let mut conj = vec.iter().sum();
+            } else {
+            let mut conj = INFINITY;    
+        }
+        conj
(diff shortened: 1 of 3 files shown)


crates/on-chain/src/objectives/basket_liquidation.rs
```diff
@@ -0,0 +1,62 @@
+use std::f64::{INFINITY, EPSILON, sqrt};
+
+/// Basket Liquidation objective function.
+pub struct BasketLiquidation {
+    /// Index token to swap into.
+    i: usize,
```
Autoparallel, line 6: what is `i`? Can we give this a better name?

crates/on-chain/src/objectives/linear_nonnegative.rs
```diff
@@ -0,0 +1,53 @@
+use std::f64::INFINITY;
+
+/// Linear Nonnegative objective function.
+pub struct LinearNonnegative {
+    /// Price vector of assets.
+    c: Vec<f64>,
```
Autoparallel, line 6: What is `c`?

Autoparallel (changes requested):

PR #250, opened 2023-05-04
Estelle/gbm unit test

### LINKED ISSUE #235: Unit test for `GBM::generate()`
Write a test in the `price_process.rs` file that does the following:

- [ ] Verifies that seeded randomness is correct, i.e., using the same seed does yield the same output and using a different seed yields a different output.
- [ ] Verifies that a single step is computed properly.

0xEstelle:
Merge #249 first! this is built off that branch. Closes #235. Wrote a unit test for the GBM completing the two checklist tasks. Namely theres a test for the seeded randomness to ensure consistency and a test for the gbm generation to ensure its being computed correctly. Leaving the OU one for a community member to do potentially


0xJepsen (approved):
Looks good to me :) i see some familiar logic from the Julia implementation.

PR #256, opened 2023-05-04
Colin/journaler

### LINKED ISSUE #254: Create a `Journaler` agent
Create an `Agent` called `Journaler` that implements the following:

- [x] Event filters that capture all instances of certain events.
- [x] Sole write access to a `.csv` file that it can echo relevant event logs into.
- [x] Writes events into `.csv`.

I plan to write a test that will look at price changes of the `liquid_exchange` over time and store these in a `.csv` in a column format.

Autoparallel:
Creates a `Journaler` agent that can write down events to a `.csv.` 

Closes #254

Added additional function to the `SimulationManager` called `shut_down()` which clears out all of the `senders` for the channels.

### DIFF
--- a/crates/bindings/src/account_lib.rs
+++ b/crates/bindings/src/account_lib.rs
@@ -7,14 +7,16 @@ pub use account_lib::*;
     clippy::upper_case_acronyms,
     clippy::type_complexity,
     dead_code,
-    non_camel_case_types,
+    non_camel_case_types
 )]
 pub mod account_lib {
     #[rustfmt::skip]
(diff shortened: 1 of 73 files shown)


0xEstelle (approved):
changes look good. just want to verify though, the journaler currently only logs and saves 1 column of data to a csv right?

0xJepsen (approved):
This looks sweet, Will be cool to use the [visualization-rs](https://github.com/primitivefinance/visualization-rs) here later down the line too!

PR #278, opened 2023-05-09
Colin/portfolio arbitrageur

### LINKED ISSUE #264: Create SimpleArbitrageur for LiquidExchange <> Portfolio
In the `portfolio_sim.rs` we need an arbitrageur that is going to properly arb the pools with the correct arb amounts. 

- [ ] Given a target price and the current pool price, compute the arbitrage amount.
- [ ] Use the auto-deployed math scripts for this.

Autoparallel:
Closes #264

Refactoring of simulations also done here.


Autoparallel:
This currently still has bugs with `InvalidInvariant` but the logic seems to be working.

0xEstelle (approved):
this is awesome!

PR #285, opened 2023-05-11
Colin/uniswap v2 sim

Autoparallel:
Working on getting more contracts in here and setting up a v2 sim


0xJepsen:
@kinrezC Figured out we weren't approving the router, there might be more wrong but that's based. 
He used a tool called [tenderly](https://tenderly.co/) and found a successful call on the main-net contract and walked through the execution trace.

0xJepsen:
Okay so Just thinking out load here, i think the approach we should go for here is to add a precompile (separate from git submodules) manually like they have here: https://www.npmjs.com/package/@uniswap/v2-core?activeTab=code
this way we don't have to add any additional language dependancies and what not. I think we can get by not needing generating the bindings for the from this bytecode. 

What this would look like a bit of a hack but :
- need to get the bytecode in the right format, in the lib it looks likes it's hex where in the bindings it's comma seperated bytecodes
- move paste in the formated bytecode into the bindings contract
- test to see if things work 
- if so, move that binding to a new `precompile` somewhere in the crate where it wont be overwritten.

Autoparallel:
This was so fucked

0xEstelle (approved):
amazing

0xEstelle:
Arbiter Math tests are failing here

Issue #290, opened 2023-05-13
`getReserves()` reverts for UniV2 initialization


Autoparallel:
In #285 there is a revert when trying to `addLiquidity` to a UniV2 pool. From what I can tell, it happens from the `.getReserves()` function called on the pair. This should return `(0,0)` initially.

0xJepsen:
is this closed with #285?

PR #296, opened 2023-05-17
Arb math for Uniswap V2 <> Liquid Exchange Sim

### LINKED ISSUE #292: Implement the equivalent `compute_arb_size` for Uniswap V2
Repeat the same arb calculator but for a constant product curve instead. This would effectively calculate the arbitrage quantity between Uniswap V2 and the infinitely liquid exchange. Will be located in `crates/cli/src/uniswap/arbitrage.rs`, mirroring structure of other sim.

### LINKED ISSUE #293: Adjust `compute_arb_size` to seek the arbitrage bound
Currently it is set up in a no arbitrage bound environment but this need to be adjusted so the price being targeted by the `compute_arb_size` function is fee adjusted.

0xEstelle:
Closes #293 and closes #292 . The is just a draft for now, I've written the compute arb logic. next step is test. I suspect a `getReserves` revert error considering issue #290 . This PR also adjusts the Portfolio arb math to account for a non-zero arb bound.


Autoparallel (approved):
lfgtm

PR #310, opened 2023-05-19
Colin/uniswap config toml

### LINKED ISSUE #302: Use `config.toml` for Uniswap simulation
Right now we have this much parsing set up for the simulations:

```rust
enum SimSubcommands {
    #[clap(about = "Runs portfolio simulation")]
    Portfolio,
    #[clap(about = "Runs Uniswap V3 simulation")]
    Uniswap,
}
#[tokio::main]
async fn main() -> Result<(), Box<dyn Error>> {
    let args = Args::parse();

    match &args.command {
        Some(Commands::Sim(sim_args)) => match &sim_args.subcommand {
            SimSubcommands::Portfolio => {
                sim::portfolio::run()?;
            }
            SimSubcommands::Uniswap => {
                sim::uniswap::run()?;
            }
        },
// ~ snip ~ //        
```

However, when we get into the simulation, we are hardcoding parameters for objects like the price path:
```rust
// Get prices
    let ou = OU::new(0.001, 50.0, 1.0);
    let price_process = PriceProcess::new(
        PriceProcessType::OU(ou),
        0.01,
        "trade".to_string(),
        10,
        1.0,
        1,
    );
    let prices = price_process.generate_price_path().1;
```

We should add the ability to change these parameters using the `config.toml` file we have set up.

Autoparallel:
Closes #302.

I cleaned up a lot here and renamed quite a bit of things. I tried to add comments where helpful and cut down on complexity in our input parsing as well. `serde` is quite helpful in doing more advanced setups in a configuration file.


0xJepsen (changes requested):
Why change sim to simulate?

0xJepsen (approved):
lgtm

PR #315, opened 2023-05-19
Colin/simulation sweeps

### LINKED ISSUE #303: Parallelize over parameter sweeps
Given #302 is completed, we can provide ranges of parameters in the `config.toml`. 

Now, it would be good to spawn multiple threads to begin working on multiple different simulations for different parameter settings.

- [ ] Add a `workers` variable to the config to set the number of simulations to run concurrently.
- [ ] Let the parameters be given by ranges of parameters, i.e., something like `1..5`. 
- [ ] Set up a nice way to output CSVs from this.

Autoparallel:
Now we can set more simulation parameters from a config. We can provide a number of `price_paths` and `number_of_volatility_steps` that we can parallelize over with a `worker_limit` enforcing a maximal number of initial workers (threads) to spawn.

Closes #303


0xJepsen (commented):
What is the .empty file?

0xEstelle (approved):
lookin clean

PR #320, opened 2023-05-22
Colin/refactor crate structure

### LINKED ISSUE #318: Refactor crate structure
We can remove `arbiter` from the crates directory and place it in the root as `src`. Then we can edit the workspace's `Cargo.toml` file to allow for the binary to be in the `./src` directory. All other crates can remain in the crates directory.

It may be best to get some name changes in now. I propose renaming the `./lib` to `./contracts` and `./crates` to `./lib` since it will house only our lib crates.

- [x] Replace `.,/crates/arbiter` with `./src`.
- [x] Rename `./lib` to `./contracts`

Autoparallel:
Merge #319 first.

This closes #318.

As a bonus I used `sed` to make sure we have the correct code hash for the v2-periphery contracts when generating bindings with our script. Also installs each set of contracts independently.


0xJepsen (approved):
Lgtm🫡

PR #329, opened 2023-05-24
Estelle/Issues

### LINKED ISSUE #316: Have arbitrageur swap against liquid exchange
We need the arbitrageur to actually arbitrage and swap against two exchanges in our simulations. Right now in the Uniswap simulation, we just have a swap on the DEX itself, and not also against the liquid exchange.

### LINKED ISSUE #325: Record all price changes not just when a swap occurs
We only record the price changes and store them when a swap/arbitrage event occurs. We should be saving all the price changes and just recording that there may be no change in the price for the Uniswap pool.

0xEstelle:
Closes #316 and closes #325. Not too big should be short and clean


0xJepsen (approved):
This looks good to me!

PR #356, opened 2023-05-30
Waylon/apply new api to sim

### LINKED ISSUE #360: Combine a lot of the Simulation / Manager log
right now this is a little fractionalized, i think we should unify this so we don't have multiple structures with simulation attributes fractionalizing the abstractions.

0xJepsen:
Tasks
- [ ] close #360


Autoparallel:
@0xJepsen can you switch this to a draft?

0xJepsen:
Not sure it makes sense to have all the CI passing yet since it's into dev. But conflicts are resolved now.

PR #362, opened 2023-06-03
Colin/evm tx receiver

### LINKED ISSUE #262: Broadcast events and serialized filtering
## Broadcasting
The way we emit events to agents now is clunky. Namely, every agent gets an `event_receiver` and there is a corresponding sender owned by the `SimulationEnvironment` like so:
```rust
pub struct SimulationEnvironment {
    /// The EVM that is used for the simulation.
    /// The sender on the event channel that is used to send events to the agents and simulation manager.
    pub(crate) event_senders: Vec<Sender<Vec<Log>>>,
}
```

Instead, we can let agents have subscribe to channels of a `tokio::sync::broadcast`.

## Serialized filtering
The filters that come with `ethers-rs` contract bindings allow for the JSON-RPC to filter out the serialized data before returning the relevant post-filter JSON data. This could be good to use in tandem while thinking about the above. The benefit here is that we can get more usage out of the contract bindings and not have to build filters in such an involved way. This could also be a minor speedup.

### LINKED ISSUE #349: Let `evm` receive txns over a channel
I am proposing that we drop the idea of moving the environment/evm around. It causes a lot of UX issues with the borrow checker and is pretty pointless. Instead, let's have the evm receive txns via a channel that all agents own a sender for. This way an agent can also not need to mention an environment when making a call since they will implicitly be attached to the right environment with a sender/receiver pair.

I plan to do this and use `tokio` channels.

### LINKED ISSUE #357: Replace `monitor_events` with `watch`
Also improve this function so it will asynchronously await for events that do pass through the filter.

### LINKED ISSUE #358: Improve `detect_arbitrage`
It should use the new `watch` function to await for arbitrage events properly. This can be made much more succinct.

### LINKED ISSUE #359: Implement event stream
We need to create an event stream object that will allow us to watch for events asynchronously.  This is how we read events from the chain with the "default" ethers middleware:
```rust
let mut event_stream = self
            .provider
            .watch(&event_filter) // stream events live
            .await
            .expect("Failed to create event stream");

        while let Some(log) = event_stream.next().await {
            println!("Event data: {:?}", log);
        }
```

This is from `onchain::monitor`. Our stream should reflect this structure exactly.

Autoparallel:
Goals are to:
- [x] Close #359 
- [x] Close #358
- [x] Close #357
- [x] Close #349 
- [x] Close #262 

Some of the above issues may need amended slightly. We need to prep for proper middleware by implementing `StreamExt` for our event broadcast and agent watcher. Once this is done we can close the remaining issues and merge this PR.

The PR #356 is forked from a commit in this branch. We should merge this prior to #356.


Autoparallel:
We need to fix the tests in `simulate/lib.rs`

Autoparallel:
Let's go!

0xJepsen (approved):
This was a bit one but i think this looks pretty clean.

Issue #375, opened 2023-06-23
Basic trait method implementations


Autoparallel:
# Problem
Given all the middleware traits (there's a lot), this is a list of the types of them that we need to function for Arbiter.

## Tech we already have
- [ ] Transaction sending and results.
- [ ] Event subscription and watching.
--- 
## New tech
- [ ] Block data.
- [ ] Balance information.
- [ ] Gas pricing.
- [ ] TxPool.
---
Both of the above are roughly ranked in order of importance to yielding MVP middleware.

Autoparallel:
Closed by #378

PR #378, opened 2023-06-23
`RevmMiddleware` and deep refactor

### LINKED ISSUE #372: Scaffold out `Middleware`
- [x] Get the necessary types for the `ethers::providers::Middleware` trait.
- [x] Get the necessary functions for the `ethers::providers::Middleware` trait.
- [x] Get the necessary imports.
- [x] Mark functions as `unimplemented!()` if we do not believe we need them (ever).
- [x] Mark functions as `todo!()` if we believe we need to implement them.

Autoparallel:
Middleware compiles now. Some is based on `ethers-reth` and some of the trait functions have been commented out.

- [x] Closes #372

---

I want to keep adding to this branch so that we do not disrupt other flow on `main`.


Autoparallel:
I plan to fork off of this now to close related issues.

Autoparallel:
The last PR #395 into here did a lot. We should list it out and link some issues @0xJepsen.

0xJepsen:
Will close #375

Autoparallel:
@Alexangelj @0xJepsen Do we need the `foundry.toml` in this repo?

Autoparallel:
@0xJepsen the book in `/documentation` is also horribly out of date now. Can we just remove the book and focus on Cargo docs for now and return to the book after we get templates and everything done?

Alexangelj:
> @Alexangelj @0xJepsen Do we need the `foundry.toml` in this repo?

I think we can make a sub crate like arbiter-contracts and have it in there, where arbiter-contracts are useful contracts to either test arbiter functionality or import into your sim, what do you think?

arbiter-core/src/middleware.rs
```diff
@@ -0,0 +1,678 @@
...
+                    let ethers_logs = revm_logs_to_ethers_logs(received_logs);
+                    for log in ethers_logs {
+                        if filtered_params.filter_address(&log)
+                            && filtered_params.filter_topics(&log)
+                        {
+                            logs.push(log);
+                        }
+                    }
+                    break;
+                }
```
0xJepsen, line 547: This loop never actually loops. Could we do something like
```rust
if let Ok(received_logs) = filter_receiver.receiver.recv() {
    let ethers_logs = revm_logs_to_ethers_logs(received_logs);
    for log in ethers_logs {
        if filtered_params.filter_address(&log)
            && filtered_params.filter_topics(&log)
        {
            logs.push(log);
        }
    }
}
```
or maybe i am not understanding things the right way?

arbiter-core/src/middleware.rs
```diff
@@ -0,0 +1,678 @@
...
+
+                // TODO: This can probably be avoided somehow
```
0xJepsen, line 549: Should we try to address this before we merge tomorrow? I am okay ice-boxing things like this till after the minor release. In some sense thats what the patches are for. In either case we should make an issue for this with more detail on what needs to be done

arbiter-core/src/middleware.rs
Autoparallel: Why doesn't it loop? Filter receiver returns a `Vec<Log>`, right? So it loops over each log in the vector which could be one, but could also be more.

arbiter-core/src/middleware.rs
Autoparallel: No I don't want to mess with this now. Icebox.

arbiter-core/src/middleware.rs
Autoparallel: We could ditch the loop and `.iter().filter()` probably if you prefer.

0xJepsen (approved):
A long time coming, really happy to see this :) Amazing work @Autoparallel Was great to be on this journey with you ✨

Autoparallel:
Amazing!!! 🥳

PR #381, opened 2023-06-25
Nit: Only print "Starting..." message on subcommand usage

Sabnock01:
Only logs `Starting Arbiter...` on subcommand usage. When no subcommands are provided, the help message is printed and no Arbiter operations are performed.


.gitmodules
```diff
@@ -28,6 +28,3 @@
 	path = contracts/solidity-lib
 	url = https://github.com/Uniswap/solidity-lib
 	branch = v2.1.0
-[submodule "contracts/vault-v2"]
```
kinrezC: We have some ideas for making submodule/contract management much more straightforward. Sorry if this is a pain point!

PR #387, opened 2023-06-29
Bug/infinity case

kinrezC:
- This sim run will hit the `Infinity()` revert case


0xJepsen (commented):
This is primarily due to too big a swap size or if swap size is zero.

PR #389, opened 2023-06-29
Waylon/init build

### LINKED ISSUE #374: arbiter `init` flag
We are imagining that the developer experience of using arbiter will follow something like this
- User installs the binary
- User calls `arbiter init` which will initialize an `arbiter` directory, with a bindings subdirectory, a simulations sub directory (maybe with an example uniswap simulation) and perhapse include a directory for arbiter specific utility contracts

This issue should play implement an MVP of this feature add the flag to our clang config in main

### LINKED ISSUE #388: Arbiter `bind` arbiter bind flag
Bind should look for contract in foundry default place
generates bindings as a module with forge bind to the auto-generated module from `arbiter init`

0xJepsen:
Closes #374 
Closes #388


bin/bind.rs
```diff
@@ -0,0 +1,26 @@
+use std::process::Command;
+
+pub(crate) fn bind_forge() -> std::io::Result<()> {
```
Autoparallel, line 3: `bind_forge` not `forge_bind`?

bin/init.rs
```diff
@@ -1,63 +1,39 @@
...
-}"#;
+    fn main() { 
+        let _ = simulations::testsim::run();
+    }"#;
 
-    // TODO
     let toml = format!(
         r#"[package]
-name = "{} Simulation"
+name = "arbitersim"
```
Autoparallel, line 15: What was wrong with the other name?

Autoparallel (commented):
Left comments. Also curious about the reasoning behind the file structure choices.

Autoparallel (approved):

PR #395, opened 2023-07-03
Waylon/experiment

0xJepsen:


0xJepsen:
I'm thinking that we can actually remove the bind script and the git submodules after we finish this refactor.

Autoparallel (approved):
Yeah we cookin'

PR #409, opened 2023-07-28
Waylon/time oracle

0xJepsen:


arbiter-core/src/environment.rs
```diff
@@ -32,63 +50,84 @@ pub enum State {
...
+    /// label for the environment
     pub label: String,
     pub(crate) state: State,
     pub(crate) evm: EVM<CacheDB<EmptyDB>>,
+    /// Connection to the environment
     pub connection: Connection,
     /// Clients (Agents) in the environment
-    pub clients: Vec<Arc<Agent<RevmMiddleware>>>,
-    // pub deployed_contracts: HashMap<String, Contract<RevmMiddleware>>,
+    pub clients: HashMap<String, Agent<RevmMiddleware>>,
```
Autoparallel, line 62: This shouldn't be necessary as all agents will be run in an environment. There's no need to reference by name.

arbiter-core/src/environment.rs
```diff
@@ -32,63 +50,84 @@ pub enum State {
...
+            lambda: 0.0,
         }
     }
-    // TODO: We need to make this the way to add agents to the environment.
-    // in `agent.rs` we have `new_simulation_agent` which should probably just be called from this function instead.
-    // OR agents can be created (without a connection?) and then added to the environment where they will gain a connection?
-    pub fn add_agent(&mut self, agent: Agent<RevmMiddleware>) {
-        self.clients.push(Arc::new(agent));
+    /// Set the expected events per block
+    pub(crate) fn configure_lambda(&mut self, lambda: f64) {
```
Autoparallel, line 102: Why do this when we can just set lambda in the constructor?

arbiter-core/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 pub mod agent;
 pub mod bindings;
+/// Module for managing the environment.
```
Autoparallel, line 7: We don't really need doc comments in the lib, instead we can use module level comments like `//!` (see here: https://doc.rust-lang.org/reference/comments.html)

arbiter-core/src/math/mod.rs
```diff
@@ -2,6 +2,7 @@
 
 use ethers::types::U256;
 
+/// Stochastic process module.
```
Autoparallel, line 5: Not really a necessary comment imo

arbiter-core/src/math/stochastic_process.rs
```diff
@@ -0,0 +1,251 @@
...
+    /// number of time steps between t_0 and t_n
+    pub n_steps: usize,
+    /// how many process trajectories to simulate
+    pub m_paths: usize,
+    /// run in parallel or not (recommended for > 1000 paths)
+    pub parallel: bool,
+}
+
+/// Create new process and run euler maruyama.
+pub fn new_procces(
```
Autoparallel, line 45: typo

arbiter-core/src/math/stochastic_process.rs
```diff
@@ -0,0 +1,251 @@
...
+            config.t_0,
+            config.t_n,
+            config.n_steps,
+            config.m_paths,
+            config.parallel,
+        ),
+    };
+
+    Ok(trajectories)
+}
```
Autoparallel, line 109: I'm not sure we need this. Can you explain why we want to have this versus just doing something like:
```rust
pub use rustquant::some_module::SomeType;
```
in order to expose the crate to our end users?

We could also do:
```rust
pub type MyType = rustquant::some_module::ComplexOrUnclearTypeName;
```

Autoparallel (changes requested):
See feedback!

arbiter-core/src/environment.rs
Autoparallel: Unless we want to be able to change it as we go through a sim? Either way, we should probably set a starting `lambda` instead of just having it be 0. 

We can also make the `lambda = 0` case one where we don't even update blocks.

arbiter-core/src/math/stochastic_process.rs
```diff
@@ -0,0 +1,251 @@
...
+
+/// Sample Poisson process.
+pub fn poisson_process(lambda: f64, length: usize) -> Result<Vec<i32>> {
```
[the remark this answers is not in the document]
0xJepsen: Yes this is correct. Do you want this to change?

arbiter-core/src/math/stochastic_process.rs
0xJepsen: We could also do that. Do you think this is better?

Autoparallel: I think it reduces our codebase and gives us the same capabilities, so I'm in favor for the reduction.

I'd probably go for the `pub use` version and when we parse a `config.toml` (eventually), we can just have `x_0`, `t_0`, `t_n`, etc. as values in the `config.toml`.

arbiter-core/src/environment.rs
0xJepsen: So I made it an option and only applied it if it is some. I kept the ability for users to configure it if they want which i think is good for now.

arbiter-core/src/environment.rs
```diff
@@ -32,63 +50,84 @@ pub enum State {
...
         //give all agents their own thread and let them start watching for their evnts
         thread::spawn(move || {
             while let Ok((to_transact, tx, sender)) = tx_receiver.recv() {
                 // Execute the transaction, echo the logs to all agents, and report the execution result to the agent who made the transaction.
+                if counter.load(Ordering::Relaxed) >= expected_occurance[0] as usize {
+                    evm.env.block.number += U256::from(1);
+                    counter.store(0, Ordering::Relaxed);
+                }
                 evm.env.tx = tx;
+                counter.fetch_add(1, Ordering::Relaxed);
```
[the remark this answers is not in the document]
0xJepsen: Nice catch

arbiter-core/src/lib.rs
0xJepsen: Nice

arbiter-core/src/manager.rs
```diff
@@ -34,6 +36,28 @@ impl SimulationManager {
         Ok(())
     }
 
+    /// Configure the lambda for an environment.
+    pub fn configure_lambda(&mut self, lambda: f64, environment_label: String) -> Result<()> {
```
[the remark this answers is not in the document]
0xJepsen: See above comment

arbiter-core/src/math/mod.rs
0xJepsen: Removed

arbiter-core/src/math/stochastic_process.rs
0xJepsen: Okay i took it out and just made it a `pub use`

arbiter-core/src/environment.rs
```diff
@@ -34,23 +51,30 @@ pub enum State {
...
 
+/// The environment struct.
 pub struct Environment {
+    /// label for the environment
     pub label: String,
     pub(crate) state: State,
     pub(crate) evm: EVM<CacheDB<EmptyDB>>,
     pub(crate) tx_sender: TxEnvSender,
-    tx_receiver: TxEnvReceiver,
+    pub tx_receiver: TxEnvReceiver,
```
Autoparallel, line 61: This shouldn't need to be pub, should it? The only one using this should be the EVM that is ran in this module. Things could go bad if an end user accesses this.

arbiter-core/src/environment.rs
```diff
@@ -101,45 +127,78 @@ impl Environment {
...
                         Ok(val) => val,
                         // URGENT: change this to a custom error
                         Err(_) => panic!("failed"),
                     };
-                    sender.send(execution_result.result).unwrap();
+                    let result_and_block = RevmResult {
+                        result: execution_result.result,
+                        block_number: convert_uint_to_u64(evm.env.block.number).unwrap(),
+                    };
+                    sender.send(result_and_block).unwrap();
```
Autoparallel, line 201: Do you have a test for this new logic? I think that would be good so we know it works as intended.

arbiter-core/src/math/stochastic_process.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub t_0: f64,
+    /// terminal time point
+    pub t_n: f64,
+    /// number of time steps between t_0 and t_n
+    pub n_steps: usize,
+    /// how many process trajectories to simulate
+    pub m_paths: usize,
+    /// run in parallel or not (recommended for > 1000 paths)
+    pub parallel: bool,
+}
```
Autoparallel, line 42: Do we need any of this?

arbiter-core/src/tests.rs
```diff
@@ -9,18 +9,19 @@ use ethers::{
...
+    middleware::{*},
 };
 
 const TEST_ARG_NAME: &str = "ArbiterToken";
 const TEST_ARG_SYMBOL: &str = "ARBT";
 const TEST_ARG_DECIMALS: u8 = 18;
 const TEST_MINT_AMOUNT: u128 = 1;
 const TEST_MINT_TO: &str = "0xf7e93cc543d97af6632c9b8864417379dba4bf15";
+const TEST_ENV_LABEL: &str = "test_env";
+const TEST_AGENT_NAME: &str = "test_agent";
```
Autoparallel, line 24: These were already sourced from `environment::tests` and `agent::tests` btw. I don't think they needed to be added here at all. The analyzer errors you're seeing are just because of the `#[cfg(test)]` that is in those modules making it so that isn't loaded in your development environment. If you try running the tests, you'll see they pass just fine.

arbiter-core/src/agent.rs
```diff
@@ -27,8 +27,9 @@ pub struct Agent<A: Attached> {
     pub behaviors: Vec<Box<dyn Behavior>>,
 }
 
+
 impl Agent<NotAttached> {
-    pub(crate) fn new<S: Into<String>>(name: S) -> Self {
+    pub fn new<S: Into<String>>(name: S) -> Self {
```
Autoparallel, line 32: Good catch

arbiter-core/Cargo.toml
```diff
@@ -1,14 +1,16 @@
...
 [package]
-name = "core"
+name = "arbiter-core"
 version = "0.3.0"
 edition = "2021"
 
 # Dependency configuration
 [dependencies]
 # For execution
+dashmap = "5.5.0"
```
Autoparallel, line 10: What is this used for?

arbiter-core/Cargo.toml
```diff
@@ -30,7 +33,7 @@ thiserror = "1.0.30"
 # url = "2.2.2"
 
 # Workspace dependencies
-ethers = { git = "https://github.com/Autoparallel/ethers-rs.git"}
+ethers = { git = "https://github.com/primitivefinance/ethers-rs.git"}
```
Autoparallel, line 36: 👍

arbiter-core/src/agent.rs
```diff
@@ -53,12 +54,8 @@ impl Agent<NotAttached> {
 
     pub fn attach_to_environment(self, environment: &mut crate::environment::Environment) {
         let middleware = RevmMiddleware::new(&self, environment);
-        let agent = Agent::<IsAttached<RevmMiddleware>> {
-            name: self.name,
-            client: Arc::new(middleware),
-            behaviors: self.behaviors,
-        };
-        environment.add_agent(agent);
+        let agent_attached = self.attach_to_client(middleware.into());
```
Autoparallel, line 57: Oh this was smart. I am curious though, why do we need the `.into()`?

Autoparallel (changes requested):
Nothing crazy to change here I think. Mostly can you just look at the comments?

Last thing: Do the same tests that pass in the current `feat/middleware` commit still ;ass here?

I think you can just run `cargo test --all --all-features`

arbiter-core/src/environment.rs
0xJepsen: This was from the merge commit so i just left it

0xJepsen: Figured you were using it

0xJepsen: Tests all pass except for the ones with the todo()

arbiter-core/src/environment.rs
0xJepsen: These would make the current call tests fail if they are not working

arbiter-core/src/math/stochastic_process.rs
0xJepsen: I'll pull it out

arbiter-core/Cargo.toml
0xJepsen: Oh we don't need anymore it is a thread safe hashmap i was playing with before i found the atomic usize

arbiter-core/src/agent.rs
0xJepsen: Otherwise it is
```
expected struct `std::sync::Arc<_>`
   found struct `middleware::RevmMiddleware`
   ```

PR #413, opened 2023-08-14
Feat/simple

0xJepsen:
This PR pulls out a lot of things we decided not to use or are unnecessary. 
- Removes the Agent functionality
- Removes unused dependancies
- Renames the strategies.rs file to signers.rs which is more verbose
- Removes un-relevant content from the root readme
- Removes the crate readme (leftover from when is was called the simulation crate)
- Sets the version of arbiter and crate to minor release three: v0.3.0
- This PR also adds back the agent address determinism feature
- This also removes some recursive error calls that were breaking functionality with stack overflows.


README.md
```diff
@@ -8,6 +8,9 @@
 [![](https://dcbadge.vercel.app/api/server/primitive?style=flat)](https://discord.gg/primitive)
 [![Twitter Badge](https://badgen.net/badge/icon/twitter?icon=twitter&label)](https://twitter.com/primitivefi)
 
+
```
0xJepsen, line 11: I am sure we can improve on this in the future, but seemed like a step in the right direction. We should probably review more thoroughly before merging.

README.md
```diff
@@ -56,17 +33,6 @@ cargo doc --workspace --no-deps --open
...
-
-In the `contracts/` directory you can add additional smart contracts or regenerate Rust bindings. Once that is done, you will want to make sure the bindings are generated in the script:
-
-```bash
-./bind.sh
-```
-You will need to add the relevant directory for your new contracts to the script above and make sure they are also handled by `forge install`. We look forward to improving upon this UX in the future.
-
-At the moment, this only builds the bindings for the contracts in the `lib/arbmod/contracts/` and `lib/portfolio/contracts`. You can of course add an additional directory of contracts in `lib/`. Just be sure to include it when you generate bindings!
-
```
0xJepsen, line 69: This is also not relevant with the new release

arbiter-core/src/environment.rs
0xJepsen, line 1: Changes in this file are removing agents

arbiter-core/src/lib.rs
0xJepsen, line 1: Removing Agents

arbiter-core/src/environment.rs
0xJepsen, line 1: Changes here remove agents

arbiter-core/src/manager.rs
0xJepsen, line 1: Remove Agents

arbiter-core/src/middleware.rs
0xJepsen, line 1: Changes here add the optional parameter to the `RevmMiddleware` Constructor that allows us to set a seed for an address.

arbiter-core/src/tests/interaction.rs
0xJepsen, line 1: Modified the `deploy_and_start()` to return the client of the deployer for the tests. The rest of the changes here are to get the tests to pass without agents.

arbiter-core/src/tests/signer.rs
0xJepsen, line 1: Renamed from strategies

arbiter-core/src/tests/interaction.rs
```diff
@@ -1,8 +1,8 @@
 use super::*;
 
 #[tokio::test]
-async fn deploy() -> Result<()> {
-    let (arbiter_token, _environment) = deploy_and_start().await?;
+async fn test_deploy() -> Result<()> {
```
Autoparallel, line 4: i think we can just call this `deploy()` -- `test_deploy()` is redundant

arbiter-core/src/tests/signer.rs
```diff
@@ -0,0 +1,31 @@
+use std::sync::Arc;
+
+use super::*;
+const TEST_AGENT_NAME: &str = "test_agent";
+#[test]
+fn simulation_agent_wallet() {
```
Autoparallel, line 6: I think we should drop any use of `Agent` for now. There are a few places I see it still in here. Consider making this change so we can use more precise terminology in the future.

arbiter-core/src/tests/mod.rs
```diff
@@ -12,7 +14,7 @@ use ethers::{
 };
 
 use crate::{
-    agent::{tests::TEST_AGENT_NAME, *},
+    // agent::{tests::TEST_AGENT_NAME, *},
```
Autoparallel, line 17: can delete

Autoparallel (changes requested):
Left small nits

Autoparallel (approved):
looks good. I might add an `<S: Into<String>>` for creating a client.

PR #417, opened 2023-08-15
Waylon/bin docs

0xJepsen:
This PR
- Adds documentation to the binary crate
- Fixes the bind and init CLI args to use the arbiter template

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,10 @@ quote = "1.0.28"
 # Errors
 thiserror = "1.0.40"
 
+#logs 
+log = "0.4.20"
+env_logger = "0.10.0"
+
 # Dependencies for the test build and development
 [dev-dependencies]
 assert_cmd = "2.0.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,6 @@ quote = "1.0.28"
 # Errors
 thiserror = "1.0.40"
 
-# Logs 
-log = "0.4.20"
-env_logger = "0.10.0"
-
 # Dependencies for the test build and development
 [dev-dependencies]
 assert_cmd = "2.0.11"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ quote = "1.0.28"
 # Errors
 thiserror = "1.0.40"
 
-#logs 
+# Logs 
 log = "0.4.20"
 env_logger = "0.10.0"
 
--- a/arbiter-core/src/environment.rs
+++ b/arbiter-core/src/environment.rs
@@ -60,9 +60,7 @@ impl Environment {
         let mut evm = EVM::new();
         let db = CacheDB::new(EmptyDB {});
         evm.database(db);
-
         let seeded_poisson = SeededPoisson::new(block_rate, seed);
-
         evm.env.cfg.limit_contract_code_size = Some(0x100000); // This is a large contract size limit, beware!
         evm.env.block.gas_limit = U256::MAX;
 
@@ -92,7 +90,7 @@ impl Environment {
 
         let mut seeded_poisson = self.seeded_poisson.clone();
 
-        let mut counter: u64 = 0;
+        let mut counter: usize = 0;
         self.state
             .store(State::Running, std::sync::atomic::Ordering::Relaxed);
         let state = Arc::clone(&self.state);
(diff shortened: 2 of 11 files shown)


bin/bind.rs
```diff
@@ -1,25 +1,39 @@
-#![warn(missing_docs, unsafe_code)]
-
+#![warn(missing_docs)]
 use std::process::Command;
 
-pub(crate) fn bind_forge() -> std::io::Result<()> {
+use log::info;
```
Autoparallel, line 4: You have to add an [`env_logger`](https://docs.rs/env_logger/latest/env_logger/) or else logs can't be displayed.

Logs might be overkill here, actually. A print to console is probably fine when it is simple and we always want a user to see it.

bin/main.rs
```diff
@@ -12,64 +22,81 @@ use thiserror::Error;
 mod bind;
 mod init;
 
+/// Represents command-line arguments passed to the `Arbiter` tool.
 #[derive(Parser)]
 #[command(name = "Arbiter")]
 #[command(version = env!("CARGO_PKG_VERSION"))]
-#[command(about = "Data analysis tool for decentralized exchanges.", long_about = None)]
+#[command(about = "Ethereum Virtual Machine Logic Simulator", long_about = None)]
```
Autoparallel, line 29: This reminds me that we should go through and look for things like this and clean up our README and Github page in general.

Cargo.toml
```diff
@@ -31,6 +31,10 @@ quote = "1.0.28"
 # Errors
 thiserror = "1.0.40"
 
+#logs
```
Autoparallel, line 34: It's a nitpick, but please format like everything else or things just look clunky

Autoparallel (changes requested):
See the comments!

I also need to go run these commands myself to see what it's like first.

bin/main.rs
0xJepsen: Yeah 100%

0xJepsen:
> See the comments!
> 
> I also need to go run these commands myself to see what it's like first.

Awesome! Yeah my recommendations are to run these in this order (and we should add these to the readme) I'll do that actually
- `arbiter init your-project-name`
- `cd your-project-name`
- `arbiter bind`
- `cargo run`

Autoparallel:
> - cd your-project-name
> - arbiter bind

Should we add a possible input for bind to be a path to the `your-project-name/` directory and take `.` as a default?

0xJepsen:
> > * cd your-project-name
> > * arbiter bind
> 
> Should we add a possible input for bind to be a path to the `your-project-name/` directory and take `.` as a default?

I think that could be good in a later feature, but i think if we add that we should add it post release.

arbiter-core/src/environment.rs
```diff
@@ -35,7 +35,7 @@ pub struct Environment {
     pub(crate) state: Arc<AtomicState>,
     pub(crate) evm: EVM<CacheDB<EmptyDB>>,
     pub(crate) socket: Socket,
-    pub seeded_poisson: SeededPoisson,
+    pub seeded_poisson: Option<SeededPoisson>,
```
Autoparallel, line 38: Whoa wait, why are we doing this in this PR? Also this is a pretty hacky solution. It does work (and I did propose it lol) but can we get this out of here? It doesn't fit the intention of the PR.

Cargo.toml
```diff
@@ -31,6 +31,10 @@ quote = "1.0.28"
 # Errors
 thiserror = "1.0.40"
 
+# Logs 
+log = "0.4.20"
+env_logger = "0.10.0"
+
```
Autoparallel, line 37: Are we using logs anywhere in the bin?

bin/init.rs
```diff
@@ -1,149 +1,62 @@
...
+
+pub(crate) fn init_project(name: &str) -> io::Result<()> {
+    let status = Command::new("git")
+        .arg("clone")
+        .arg("https://github.com/primitivefinance/arbiter-template.git")
+        .arg(name)
+        .status()?;
+
+    if !status.success() {
+        error!("Failed to clone the repository.");
```
Autoparallel, line 35: Here, we are still logging.

Autoparallel (changes requested):

Autoparallel (approved):

PR #422, opened 2023-08-16
Test cleanup

Autoparallel:
# Overview of tasks

- [x] Added more test coverage focused on bindings.
- [x] Fixed a bug that arose in propagating `RevmMiddlewareError`.
- [x] Edited the `Revert` and `Halt` error variants.

### DIFF
--- a/arbiter-core/src/environment.rs
+++ b/arbiter-core/src/environment.rs
@@ -129,7 +129,7 @@ pub struct Environment {
 
     /// The [`EVM`] that is used as an execution environment and database for
     /// calls and transactions.
-    pub(crate) evm: EVM<CacheDB<EmptyDB>>,
+    evm: EVM<CacheDB<EmptyDB>>,
 
     /// This gives a means of letting the "outside world" connect to the
     /// [`Environment`] so that users (or agents) may send and receive data from
--- a/arbiter-core/src/middleware.rs
+++ b/arbiter-core/src/middleware.rs
@@ -145,17 +145,23 @@ pub enum RevmMiddlewareError {
 
     /// The execution of a transaction was reverted, indicating that the
     /// transaction was not successful.
-    #[error("execution failed to succeed due to revert! output is: {cause}")]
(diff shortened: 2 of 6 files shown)


arbiter-core/src/environment.rs
```diff
@@ -129,7 +129,7 @@ pub struct Environment {
 
     /// The [`EVM`] that is used as an execution environment and database for
     /// calls and transactions.
-    pub(crate) evm: EVM<CacheDB<EmptyDB>>,
+    evm: EVM<CacheDB<EmptyDB>>,
```
0xJepsen, line 132: Remind me why we are doing this again.

arbiter-core/src/middleware.rs
```diff
@@ -167,7 +173,7 @@ impl MiddlewareError for RevmMiddlewareError {
     }
 
     fn as_inner(&self) -> Option<&Self::Inner> {
-        Some(self)
+        None
```
0xJepsen, line 176: Can you explain this to me? The middleware onion has been a little black boxed for to only handle it on one layer at a time.

arbiter-core/src/middleware.rs
```diff
@@ -308,7 +313,13 @@ impl Middleware for RevmMiddleware {
             _gas_refunded: _,
             logs,
             output,
-        } = unpack_execution_result(revm_result.result)?;
+        } = match unpack_execution_result(revm_result.result) {
```
0xJepsen, line 316: Nice i was wondering about this when i was debugging my issue

0xJepsen (approved):
Looks good to me!

arbiter-core/src/middleware.rs
Autoparallel: Oh shit. I don't actually want to do it this way. Doing it with `?` is better.

arbiter-core/src/environment.rs
Autoparallel: Because we don't actually reference the `evm` field of `Environment` anywhere else in the crate, so we are safe to keep its visibility to private. I feel it's usually good to minimize visibility to the lowest point.

PR #424, opened 2023-08-16
Waylon/tests

0xJepsen:
This PR fixes a bug we had with transactions hanging perpetually on a paused environment. This PR adds checks to both the `call` and `send_transaction` method on the middleware to reduce calls that get through. It also added proper error handling on the off chance that some get through they will return an environment error alerting the user that the environment is paused. Some predominant changes for this improvement include:
- changing all atomic ordering rules to `SeqCst` to reduce concurrency bugs  
- Adds a `TransactionOutcome` enum to the `RevmResult` to indicate environment errors to the middleware
- Added additional error checking in the `send_transaction` and `call` implementations. 
- When we enter a paused state in the environments run method, we loop through all potentially caught last transactions and propagate a pause error for each.  
- There are some formatting and clippy changes from `cargo fmt` (trying to converge to a happy clippy before release)


arbiter-core/src/environment.rs
```diff
@@ -295,7 +295,31 @@ impl Environment {
...
+                                        cause: format!("{:?}", e),
+                                    },
+                                )?,
+                            };
+                            sender.send(revm_result).map_err(|e| {
+                                EnvironmentError::Communication {
+                                    cause: format!("{:?}", e),
+                                }
+                            })?;
+                        }
```
Autoparallel, line 320: This should work I think!

Autoparallel: Just double checking the logic:
1. We set the state to paused in some way via the `AtomicState`.
2. This eventually hits the `send_transaction()` and `call()` function conditional checks and will eventually error out the transactions/calls that users are trying to send.
3. All the while we also eventually end up in this `Paused` state and enter the `while let` on the `try_recv()` and consume all the possible txs in the queue.
4. Now we rest easy until the condvar is signaled to restart the environment.

Sound right?

arbiter-core/src/environment.rs
```diff
@@ -447,13 +471,38 @@ pub(crate) struct Socket {
...
+    ///
+    /// Contains the result of the transaction's execution.
+    Success(ExecutionResult),
+
+    /// Represents a failed transaction due to some error.
+    ///
+    /// Contains information about the error that caused the transaction
+    /// failure.
+    Error(EnvironmentError),
+}
```
Autoparallel, line 498: For whatever reason I still feel like we can get by without this entirely but I'm not worried about it for now.

arbiter-core/src/environment.rs
```diff
@@ -516,15 +565,15 @@ pub(crate) mod tests {
...
+        let state = environment.state.load(std::sync::atomic::Ordering::SeqCst);
         assert_eq!(state, State::Initialization);
     }
 
     #[test]
     fn run() {
         let mut environment = Environment::new(TEST_ENV_LABEL.to_string(), 1.0, 1);
         environment.run();
-        let state = environment.state.load(std::sync::atomic::Ordering::Relaxed);
+        let state = environment.state.load(std::sync::atomic::Ordering::SeqCst);
```
Autoparallel, line 576: Just read about the differences and I think this is a good call.

arbiter-core/src/manager.rs
```diff
@@ -188,30 +193,31 @@ impl Manager {
...
-                State::Running => Err(ManagerError::EnvironmentAlreadyRunning {
-                    label: environment_label.into(),
-                }),
-                State::Stopped => Err(ManagerError::EnvironmentStopped {
-                    label: environment_label.into(),
-                }),
-            },
+            }
             None => Err(ManagerError::EnvironmentDoesNotExist {
                 label: environment_label.into(),
```
Autoparallel, line 222: We can probably avoid this double match nesting which would be easier to read in the future.

arbiter-core/src/middleware.rs
```diff
@@ -549,10 +604,7 @@ impl JsonRpcClient for Connection {
                             logs.push(log);
                         }
                     }
-                    break;
                 }
-
-                // TODO: This can probably be avoided somehow
```
Autoparallel, line 555: Okay a nitpick but I think this TODO is valid. Why did you delete it?

arbiter-core/src/tests/interaction.rs
```diff
@@ -253,3 +254,42 @@ async fn transaction_loop() -> Result<()> {
     }
     Ok(())
 }
+
+#[tokio::test]
+async fn test_pause_prevents_processing_transactions() {
```
Autoparallel, line 259: don't need to have `test_` in the name of a test, it's redundant.

arbiter-core/src/environment.rs
0xJepsen: Thats exactly right!

arbiter-core/src/environment.rs
0xJepsen: We can change it in the future, if we want to propagate the Environment errors to the client we will need something like this.

arbiter-core/src/middleware.rs
0xJepsen: I agree i think it is still valid. Do you think we can we make an issue for it instead of leaving it in so that the code base is a little cleaner?

arbiter-core/src/tests/interaction.rs
0xJepsen: I know it's not necessary but it helps me and allows me to be precise when naming tests. If it bothers you we should chat more about it and maybe make changes to a style guide.

arbiter-core/src/middleware.rs
Autoparallel: I don't mind a `TODO` here or there in a codebase. But we should also make an issue for it, yeah.

PR #425, opened 2023-08-16
Fix CI/CD

0xJepsen:
This PR reflects some minimal changes to our CI/CD
- CI/CD now on runs on PRs into Main 
- Dependabot now wont check for our old crates and will check for our new one `arbiter-core`


.github/workflows/bump_version.yml
Autoparallel, line 1: Hey, can we actually just icebox this entirely? It was never really working correctly before.

Autoparallel (changes requested):
One idea to look at. Everything else looks great.

.github/workflows/bump_version.yml
0xJepsen: Yeah i'm totally down with that

Autoparallel (approved):

PR #443, opened 2023-08-17
Bug/filter test termination

### LINKED ISSUE #423: bug: `filter_address()` and `filter_topics()` tests hang forever
# Issue

The tests should not hang forever but should timeout after the second event is filtered through and hence does not appear. The fact that the test is hanging means the behavior of `arbiter-core` is as intended, but the test itself is built improperly.

---

## Detail:

Take a look at this test `tests::interaction::filter_address` as it is boilerplate for the other:
```rust
#[tokio::test]
async fn filter_address() -> Result<()> {
    let (arbiter_token, _environment, client) = deploy_and_start().await.unwrap();
    let mut default_watcher = client.watch(&Filter::default()).await?;
    let mut address_watcher = client
        .watch(&Filter::new().address(arbiter_token.address()))
        .await?;

    // Check that both watchers get this event
    let approval = arbiter_token.approve(
        client.default_sender().unwrap(),
        ethers::types::U256::from(TEST_APPROVAL_AMOUNT),
    );
    approval.send().await?.await?;
    let default_watcher_event = default_watcher.next().await.unwrap();
    let address_watcher_event = address_watcher.next().await.unwrap();
    assert!(!default_watcher_event.data.is_empty());
    assert!(!address_watcher_event.data.is_empty());
    assert_eq!(default_watcher_event, address_watcher_event);

    // Create a new token contract to check that the address watcher only gets
    // events from the correct contract Check that only the default watcher gets
    // this event
    let arbiter_token2 = ArbiterToken::deploy(
        client.clone(),
        (
            format!("new_{}", TEST_ARG_NAME),
            format!("new_{}", TEST_ARG_SYMBOL),
            TEST_ARG_DECIMALS,
        ),
    )?
    .send()
    .await?;
    let mint2 = arbiter_token2.mint(
        ethers::types::H160::from_str(TEST_MINT_TO)?,
        ethers::types::U256::from(TEST_MINT_AMOUNT),
    );
    mint2.send().await?.await?;
    let default_watcher_event = default_watcher.next().await.unwrap();
    assert!(!default_watcher_event.data.is_empty());
    println!("default_watcher_event: {:#?}", default_watcher_event);

    // Use tokio::time::timeout to await the approval_watcher for a specific
    // duration The timeout is needed because the approval_watcher is a stream
    // that will never end when the test is passing
    let timeout_duration = tokio::time::Duration::from_secs(1); // Adjust the duration as needed
    let timeout = tokio::time::timeout(timeout_duration, address_watcher.next());
    match timeout.await {
        Result::Ok(Some(_)) => {
            // Event received
            panic!("This means the test is failing! The filter did not work.");
        }
        Result::Ok(None) => {
            // Timeout occurred, no event received
            println!("Expected result. The filter worked.")
        }
        Err(_) => {
            // Timer error (shouldn't happen in normal conditions)
            panic!("Timer error!")
        }
    }
    Ok(())
}
```

`address_watcher.next()` should **never** receive a new event, and hence we need to wait for some delay and after the delay, time the test out to signal it passing.

0xJepsen:
This PR Closes #423 by fixing our filter watcher watcher tests that were initially commented out here:
https://github.com/primitivefinance/arbiter/blob/bb4b7af1d985739751de385336f7bdfe9058610a/arbiter-core/src/tests/interaction.rs#L104-L218

They now pass and terminate with the expected behavior. This was done by manually polling the stream and matching on the poll.

### DIFF
--- a/arbiter-core/src/tests/interaction.rs
+++ b/arbiter-core/src/tests/interaction.rs
@@ -1,3 +1,12 @@
+use std::{
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use anyhow::Ok;
+use assert_matches::assert_matches;
+use futures::stream::Stream;
+
 use super::*;
 use crate::bindings::arbiter_math::ArbiterMath;
 
@@ -58,6 +67,17 @@ async fn transact() -> Result<()> {
 }
 
 #[tokio::test]
+async fn filter_id() -> Result<()> {
+    let (arbiter_token, _environment, client) = deploy_and_start().await.unwrap();
+    let filter_watcher_1 = client.watch(&Filter::default()).await?;
+    let filter_watcher_2 = client
+        .watch(&Filter::new().address(arbiter_token.address()))
+        .await?;
+    assert_ne!(filter_watcher_1.id, filter_watcher_2.id);
+    Ok(())
+}
+
+#[tokio::test]
 async fn filter_watcher() -> Result<()> {
     let (arbiter_token, _environment, client) = deploy_and_start().await.unwrap();
     let mut filter_watcher = client.watch(&Filter::default()).await?;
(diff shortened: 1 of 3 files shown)


arbiter-core/src/tests/interaction.rs
```diff
@@ -101,121 +121,106 @@ async fn filter_watcher() -> Result<()> {
...
+    // check that the address_watcher has not received any events
+    let mut ctx = Context::from_waker(futures::task::noop_waker_ref());
+    let poll_result = Pin::new(&mut address_watcher).poll_next(&mut ctx);
+    match poll_result {
+        Poll::Ready(Some(_event)) => panic!("Event received unexpectedly!"),
+        Poll::Ready(None) => println!("Stream completed!"),
+        Poll::Pending => println!("No event ready yet, as expected."),
+    }
+    assert_matches!(poll_result, Poll::Pending);
+    Ok(())
```
Autoparallel, line 181: Okay this is the core idea that lets this all work. Can you explain it to me?

arbiter-core/src/tests/interaction.rs
```diff
@@ -57,6 +66,17 @@ async fn transact() -> Result<()> {
...
+#[tokio::test]
+async fn filter_id() -> Result<()> {
+    let (arbiter_token, _environment, client) = deploy_and_start().await.unwrap();
+    let filter_watcher_1 = client.watch(&Filter::default()).await?;
+    let filter_watcher_2 = client
+        .watch(&Filter::new().address(arbiter_token.address()))
+        .await?;
+    assert_ne!(filter_watcher_1.id, filter_watcher_2.id);
+    Ok(())
+}
```
Autoparallel, line 78: Love to see more tests!!!

Autoparallel: MORE GREEN THINGS

Autoparallel (commented):
Please help me resolve my comments!

arbiter-core/src/tests/interaction.rs
0xJepsen: yeah so the call to `.next().await()` will not normally terminating when in the `Poll::Pending` state so what i did was bring the poll result into scope and match on the poll result so that we can terminate the tests. 
- To poll a stream manually, you need a context (Context). The context, among other things, contains a "waker", which is an object that allows asynchronous tasks to notify when they're ready to make progress.
- A "noop waker" is a waker that does nothing when woken up. Since we're only interested in the current state of the stream and don't need to actually schedule any async tasks, a noop waker is sufficient.

The subsequent match statement handles each potential result from the poll_next:
- If an event is ready, which wasn't expected in this scenario, it panics.
- If the stream is complete, it prints that.
- If there's no event ready yet (i.e., it's still waiting for an event), it prints that.

By manually polling the stream, we are able to gain direct insight into the state of the address_watcher. Instead of using .await and blocking the task indefinitely.

Autoparallel: Nice! I think that makes sense to me. The context part is the most fuzzy though.

Autoparallel (approved):
lgtm

PR #446, opened 2023-08-17
config for codecov

### LINKED ISSUE #438: ci/cd: remove codecov/patch and codecov/repo but keep upload to codecov.io
# Problem

The codecov/patch and codecov/repo CI/CD checks that run are going to fail quite often and we are not currently being this restrictive on testing. Last time we tried to remove these checks, we ended up not uploading our data to codecov.io and so we weren't seeing our badge update. We need to still get the badge update, but not let these CI/CD checks occur, if possible.

0xJepsen:
This closes #438
Waiting for checks to pass then will mark as ready for review.
The changes are with the edition of a `codecov.yaml` file that overrides the default configuration (which will fail if any pr decreases by coverage by any amount) The file has a `threshold` value you can set which specifies the amount of change in a pr (up or down) in coverage that will be tolerated. I set it to 1% with a full coverage target of 50%.


.github/codecov.yaml
```diff
@@ -0,0 +1,12 @@
...
+  status:
+    project:
+      default:
+        target: 50%
+        threshold: 1%  
+    patch:
+      default:
+        target: 50%
+        threshold: 1%
+        informational: true
```
0xJepsen, line 11: This also ensures the patch coverages are strictly informational

Autoparallel: This is great!

Autoparallel (approved):
thank you for getting this chore done

PR #451, opened 2023-08-20
feat: Add `EnvironmentParameters` struct for improved environment initialization

### LINKED ISSUE #449: feat: add `EnvironmentParameters` struct for initializing environments
# Idea
Right now we pass in a few variables when initializing an environment via the manager:
```rust
let mut manager = Manager::new();
    manager.add_environment(ENV_LABEL, 1.0, 1)?;
```
The label is fine to stay where it is, but the latter two parameters `block_rate` and `seed` could be wrapped into a struct. We will most likely have more parameters in the future (e.g., `block_time`, `manual_mine`, options on these, etc.) that we use to configure an environment, so we could start on this now.

This will be a QoL improvement and help with externally configuring environments and parameter sweeps.

## Solution
Let's create a struct like:
```rust
pub struct EnvironmentParameters {
    block_rate: f64,
    seed: u64,
}
```
and change the constructor `Environment::new()` as well as `Manager::add_environment()`.

## Alternative approaches
I'm not sure of alternate approaches, but I'm all ears.

## Additional context
The additional context here is adding some more environment parameters in the future, some of which I listed above already.

alikonuk1:
Introduced a new struct EnvironmentParameters to streamline the process of initializing environments in the Manager. This refactoring effort encapsulates block_rate and seed parameters within the new struct, making the code more readable and scalable for future additions.

Closes #449

### DIFF
--- a/arbiter-core/src/environment.rs
+++ b/arbiter-core/src/environment.rs
@@ -157,7 +157,7 @@ pub struct EnvironmentParameters {
     pub block_rate: f64,
 
     /// A value chosen to generate randomly chosen block sizes
-    /// for the environment..
+    /// for the environment.
     pub seed: u64,
 }
 
--- a/arbiter-core/src/environment.rs
+++ b/arbiter-core/src/environment.rs
@@ -151,10 +151,13 @@ pub struct Environment {
 /// This structure holds configuration details or other parameters that might
 /// be required when instantiating or updating an `Environment`.
 pub struct EnvironmentParameters {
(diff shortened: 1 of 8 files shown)


Autoparallel:
Looks like we gotta get the tests passing. @alikonuk1 do you need any help with that?

0xJepsen (approved):
Nice work! This looks good to me! If you like, it would be good to make a little pr to the template repository to reflect the changes!

PR #452, opened 2023-08-20
test: Implement `signer_collision` Test

### LINKED ISSUE #429: test: signer collision
There is a todo on an agent signer collision test. The test should just ensure that addresses generated with the same seed are the same. The test should essentially complete this todo:
https://github.com/primitivefinance/arbiter/blob/1fb8a5f4cebac831abf4350c456ffae9a7ff0378/arbiter-core/src/tests/signer.rs#L26-L31

You can make it almost identical to the above test:
https://github.com/primitivefinance/arbiter/blob/1fb8a5f4cebac831abf4350c456ffae9a7ff0378/arbiter-core/src/tests/signer.rs#L18-L24
except you want to change the seed to be the same and assert that they are the same

alikonuk1:
Implemented the `signer_collision` test in the `arbiter-core/src/tests/signer.rs` file to ensure that two clients initialized with the same seed have identical default sender addresses.

This PR closes #429


0xJepsen (approved):
lgtm! Can we get CI to pass?

PR #460, opened 2023-08-23
bench: created initial benchmarks against Anvil

### LINKED ISSUE #440: bench: run comparisons against other programs
# Problem

We should compare Arbiter and the `RevmMiddleware` to other implementations carefully such as Foundry and Anvil. The exact comparisons we can discuss further.

Autoparallel:
Created a `benches/` crate that tests the speed difference between Arbiter and Anvil over a few different scenarios. Added the output to the README.

Closes #440


benches/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+    // Set up logging.
+    if std::env::var("RUST_LOG").is_err() {
+        std::env::set_var("RUST_LOG", "error");
+    }
+    env_logger::init();
+
+    // Get input argument.
+    let args: Vec<String> = std::env::args().collect();
+    let args = args.get(1).unwrap().as_str();
+    let mut durations = vec![];
```
kinrezC, line 54:
```suggestion
    let mut durations = Vec::with_capacity(NUM_BENCH_ITERATIONS);
```

kinrezC (approved):

PR #473, opened 2023-08-30
Feat(arbiter core)/v0.5.0

### LINKED ISSUE #418: feat: add in `BlockType` enum for controlling block number during emulation
# Idea
Updated from a comment below:

I think we should externalize this more and make it more than just an Option. Perhaps an enum for this along the lines of:
```rust
pub enum BlockType {
     UserControlled,
     RandomlySampled(f64, u32),
}
```
- `UserControlled`: will give the user the ability to make calls such as `client.mine_block()` and `client.set_block_timestamp()`.
- `Randomized(f64, u32)`: would give the user the the Poisson chosen block sizes with the block rate via the `f64` inner and we would let the block times update via the `u32` inner.
- `None`: The user would not update the blocks ever.

This `BlockType` can go into the `EnvironmentParameters`.

## Other details

We should probably try to implement this in some sort of RPC-like way such as listed in the [Anvil documentation](https://book.getfoundry.sh/reference/anvil/). For example, we can implement the request:
```
"evm_mine"
```
We should link this issue to #464.

Since blocks are not actually mined *per se*, we can maybe just implement a method on the `RevmMiddleware` such as `next_block` so that any clients may access this.

---
~~For those who will not need this feature, having the block number increment is just another chance for a bug a bit of overhead. Also, it relies on the user knowing what the block rate setting is doing.~~

~~- [ ] Add in an option for the `rate_parameter` when creating an environment.~~

### LINKED ISSUE #464: feat: add ability to control block timestamps
# Idea
Some protocols such as RMM-01 depend on time implicitly, so we need to be able to have a predictable way to move time forward. This network time can be move forward solely via block timestamps, which can also be updated by updating block numbers.

Yes, without the ability to have track of time, we cannot have certain protocols run properly. 

## Solution
We should provide the user a means of control over each environment's emulator state. For instance, the block number and block timestamp. We can do so via some externalized API that allows a user to choose their methodology for controlling timestamping.

This should be built alongside #418. Looking at the [Anvil documentation](https://book.getfoundry.sh/reference/anvil/), we will want an RPC-like request here that handles one or both of the following:
```
"evm_increaseTime"
"evm_setNextBlockTimestamp"
```

At the very least, for the `BlockType::RandomlySampled` we can implement this naturally:
- [x] Increment the `evm.env.block.timestamp` by the `block_time` every time we increment to the next block.

## Alternative approaches
Alternatively, we can just have a time increment number and increment block times via this number each time we push a new block using the current method from a Poisson distribution. However, this is not ideal -- we need to provide more ways for a user to control this (or **not** control it).

## Additional context
This should be linked intimately with issues such as #418

### LINKED ISSUE #470: feat: add clone, debug, deserialize, and serialize for `EnvironmentParameters` and put label inside
# Idea
These are useful when we need to generate config files. Also, the label can go into this as well.

I am trying to generate a config file now, and this is blocking me!

## Solution
Use the relevant derive macros for this. There should be no issues.

### LINKED ISSUE #471: feat: refactor `EnvironmentParameters` to include label
# Idea
We may as well place the label inside of the `EnvironmentParameters` so that every bit of description of an environment is placed there. We still want to reference the `Environment` by this label using the `Manager`

Yes, when working on setting up simulations, I found that this provided a much better UX in setting up environments.

## Solution
Just move the label into the `EnvironmentParameters` struct.

### LINKED ISSUE #474: feat: bump version for `RustQuant` to include "seedable" feature
# Idea
Having seeded randomness is immensely helpful for debugging and replaying simulations with interesting phenomenon. We recently got this feature added to `RustQuant` via [this PR](https://github.com/avhz/RustQuant/pull/86) so we should have this now included in `arbiter-core`.

Yes. I have been writing simulations where being able to replay specific paths would be immensely helpful.

## Solution
We just use the "seedable" flag with the `RustQuant` re-export.

### LINKED ISSUE #477: chore: clean up `EnvironmentErrors`
# Idea
The errors in the `Environment` can be made more easily using `#[from]` we should do so so we can cut down on some code and redundancy.

Not really, this is just QOL.

## Solution
Use the `#[from] proc macro to map errors instead of converting all to strings.

### LINKED ISSUE #479: feat: send gas cost in tx receipt
# Idea

I would like to be able to get information about the gas consumed during the transaction.

## Solution

Seems like this value is discarded here:
https://github.com/primitivefinance/arbiter/blob/e0ccdc3f641f4227d5d201984314dc8aa2678d45/arbiter-core/src/middleware.rs#L318

## Additional

Btw seem like gas counting can be switched off completely in REVM:
https://github.com/bluealloy/revm/blob/a95a2989589c831a797c42556f39e7ad2f265ecb/crates/revm/Cargo.toml#L55

I think it might improve the speed a bit if the gas cost values ​​are **not** needed.

I tried to check the difference (using `cargo build --release`) in my case and  included this in my `Cargo.toml`:

```
revm = { version = "3.3.0" , features = ["default"]}
```
I have estimation around 00:08:00 (hh, mm, ss)

And with:
```
revm = { version = "3.3.0", features = ["default", "no_gas_measuring"] }
```
I have estimation around 00:06:00 (hh, mm, ss)

But here I'm not sure, because I'm not very familiar with Rust.

Autoparallel:
# `arbiter-core` v0.5.0
This PR will close all issues related to the next minor release for the `arbiter-core` crate. Issues are tracked with a label and [v0.5.0 milestone](https://github.com/primitivefinance/arbiter/milestone/12).

- Closes #470 
- Closes #471
- Closes #474 
- Closes #477 
- Closes #418 
- Closes #464 
- Closes #479


Autoparallel:
We should make sure to edit the template! @0xJepsen

0xJepsen (approved):
Merging, as the CI tests are all passing locally.

PR #483, opened 2023-09-08
Feat/reciepts

### LINKED ISSUE #479: feat: send gas cost in tx receipt

0xJepsen:
This PR populates the transaction receipts with full information. 

closes #479 

This PR will also get the tests to pass in the v0.5.0 feature branch.

I noticed two things while working on this. Particularly when working on the transaction hash. 
- The tx hash of course depends on rlp encodings of a collection of fields in the tx, there are some nice abstractions for this that i was looking at in this [reth crate](https://github.com/paradigmxyz/reth/blob/main/crates/rlp/src/encode.rs) that i think we should use in the future. However right now it is not implemented for revm and ethers types (pending alloy migrations) and i think the current implementation is sufficient. 
- The encoding does depend on the nonce which we aren't incrementing or bookkeeping for the clients. I think the way to go would be to implement the [nonce manager](https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/nonce_manager.rs) here

### DIFF
--- a/arbiter-core/Cargo.toml
+++ b/arbiter-core/Cargo.toml
@@ -9,6 +9,8 @@ edition = "2021"
 # Ethereum and EVM
 ethers = "2.0.8"
 revm = "3.3.0"
+reth-rlp = { git = "https://github.com/paradigmxyz/reth.git", package = "reth-rlp" }
+rlp = "0.5.2"
 
 # Serialization
 bytes = "1.4.0"
--- a/arbiter-core/Cargo.toml
+++ b/arbiter-core/Cargo.toml
@@ -9,8 +9,6 @@ edition = "2021"
 # Ethereum and EVM
 ethers = "2.0.8"
 revm = "3.3.0"
-reth-rlp = { git = "https://github.com/paradigmxyz/reth.git", package = "reth-rlp" }
-rlp = "0.5.2"
 
 # Serialization
 bytes = "1.4.0"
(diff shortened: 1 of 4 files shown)


Autoparallel (changes requested):
We're missing something here as far as I can tell. In `environment.rs` lines 346-366 we should be reseting the transaction index and gas count.

0xJepsen:
> 

Nice thought. I pushed changes to address this!

Autoparallel (approved):

PR #486, opened 2023-09-12
feat/gas-settings-and-more

### LINKED ISSUE #241: feat: adjustable gas price
# Idea

To make simulations as realistic as possible, we need to give agents some raw ether and force them to pay gas for transactions. At the moment, the gas price has been fixed to zero.

## Solution
We can set another parameter in `EnvironmentParameters` similar to how I commented on #418. We can have an enum, or something, that allows the user to choose some gas rules for the environment. They can be externalized, stochastic, or off, e.g.,:
```rust
pub enum GasRules {
     UserControlled,
     Randomized(f64),
     None(u128),
}
```
- `UserControlled`: will give the user the ability to make calls such as `client.set_gas_price()`.
- `Randomized`: would give the user gas price based on something like EIP-1559 rules. Could even play off of the `BlockType` enum I mentioned in #418. 
- `None`: The user would not change gas price ever and have it set to the internal `u128`.

This `GasRules` can go into the `EnvironmentParameters`.

We can also perhaps add functionality for gas tips later on if we expand upon an enum.

--- 

**IT IS WORTH NOTING THAT ISSUE #410 IS RELEVANT HERE TOO**

Autoparallel:
This was a mighty one. Realistically, I sought out just to implement some gas settings to allow clients to pay gas *a-la* #241. However, I did a lot more. Here is a rough list:
- Added `GasSettings` enum and placed it in `EnvironmentParameters`. Force that `GasSettings::RandomlySampled` must be used along with `BlockType::RandomlySampled`.
- Implemented some extra middleware functionality for clients:
    - `get_block_number()` -- returns the current block number
    - `get_block_timestamp()` -- returns the current timestamp
    - `get_gas_price()` -- returns the current gas price, and is used in settings the `gas_price` field of the `TxEnv` passed to the `Environment`.
    - `get_balance()` -- returns the current balance
    - `deal()`
- Had to add in some other `Instruction` variants.
    - `AddUser` -- called when creating a new client (middleware). This way they can be given a balance and eventually have cheatcodes act on them.
    - `Query` -- so that we can query certain attributes of the `Environment`, e.g., ones listed above.
- Tests for everything I can think of and a bit of a refactor of the tests module.

I'm sure the list isn't exhaustive but I'm exhausted!

I did not run any linter checks and I am missing docs. This is immaterial and we can do this in the last commits into the feat/version branch. Let's go!

Closes #241


Alexangelj (approved):
LGTM

one question: the instruction sender is a specific thing, why is that? Just curious since I havn't seen it until now

Autoparallel:
@Alexangelj Do you mean that it is aliased?

0xJepsen (approved):
nice this lgtm!

PR #506, opened 2023-09-18
feat(contracts): arbiter token auto approve on transfers

Autoparallel:
This PR adds the token spend approval upon `transfer` and `transferFrom` to cut down on additional approvals when using these tokens.

N/A


0xJepsen:
What if there are bugs in a the way that some contracts are implementing approvals? do you think this is any edge case that would be good to have them separate for?

0xJepsen (approved):
This looks good to me

Autoparallel:
> What if there are bugs in a the way that some contracts are implementing approvals? do you think this is any edge case that would be good to have them separate for?

That's fair. Maybe it's better to just have the liquid exchange properly handle approvals instead then and not change in the token logic itself. Let me close this one.

Autoparallel:
@jepsen, we might want to consider reopening this as you can't do the approvals in the exchanges.

PR #507, opened 2023-09-18
Feat(arbiter-derive): create `#[derive(Deploy)]`

Autoparallel:
Created a new crate called `arbiter-derive` which is set to be `proc_macro = true`. Added a single proc macro for `#[derive(Deploy)]` that should assist end users in having an slightly easier contract deploy process.

N/A


0xJepsen (approved):
lgtm!

PR #514, opened 2023-09-20
Feat/no git

### LINKED ISSUE #502: feat(arbiter-bin): create a project that is **not** a git repo
# Idea
If a user wants to create a new Arbiter project inside of their current smart contract repository, it may not be great to give them another whole git repository inside of it. Having a flag to disable git would be nice so you could do 
```sh
arbiter init my_project --no-git
```

Yes, we have been given a little bit of grief that this makes the project a submodule and is annoying!

## Solution
We can even just remove the `.git` directory if we would like when we clone the template.

## Alternative approaches
Don't clone and don't use `cargo new` either.

0xJepsen:
- [x] adds an optional flag `no-git` that when following the -o flag will initialize a repository without a .git dir

closes #502

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,10 @@ Arbiter treats Rust smart-contract bindings as first-class citizens. The contrac
 `arbiter bind` wraps `forge` with some convenience features that will generate all your bindings to src/biindings as a rust module. 
 [Foundry](https://github.com/foundry-rs/foundry) power-users are welcome to use `forge` directly.
 
+### Optional Arguments
+
+You can run `arbiter init -o` followed by any optional arguments. Currently the only optional argument is `no-git` that will be sure to remove the `.git` directory from the template upon initialization.
+
 ## Documentation
 
 To see the documentation for Arbiter, after cloning the repo, you can run:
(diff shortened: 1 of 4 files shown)


Autoparallel (commented):
Can you add a bit on how to use this in the readme or something?

README.md
```diff
@@ -86,6 +86,10 @@ Arbiter treats Rust smart-contract bindings as first-class citizens. The contrac
 `arbiter bind` wraps `forge` with some convenience features that will generate all your bindings to src/biindings as a rust module. 
 [Foundry](https://github.com/foundry-rs/foundry) power-users are welcome to use `forge` directly.
 
+### Optional Arguments
+
+You can run `arbiter init -o` followed by any optional arguments. Currently the only optional argument is `no-git` that will be sure to remove the `.git` directory from the template upon initialization.
```
Autoparallel, line 91: Can we make it so you can just run `arbiter init -no-git` as opposed to `arbiter init -o -no-git`?

Autoparallel (changes requested):

README.md
0xJepsen: Yes! Just pushed these changes now

Autoparallel (approved):

PR #516, opened 2023-09-21
feat(arbiter core): event capture

### LINKED ISSUE #515: feat: general purpose event logger
# Idea
In running simulations, it would be extremely handy to have a concise way of setting up events that you would like to log to some output files (probably just CSV for now). 

Yes, we should not have to write a bespoke data output collector for simulations.

## Solution
A solution is to set up a structure like so:
```rust
pub struct EventCapture<F: EthEvent + EthLogDecode + Debug + 'static> {
    event: Event<Arc<RevmMiddleware>, RevmMiddleware, F>,
    capture: Vec<BTreeMap<String, Value>>,
    pub state: Arc<AtomicBool>,
}
```
This struct can then have a run function that can be called to start the event logging to move the `state` into `true` to represent it running and capturing events. Stopping the environment can be a means to control the shut down of this logger.

## Alternative approaches
We could instead just store the states in a DB on disk and have people post process from this on their own. But I'd like to give the above a shot nonetheless as optionality is good.

Autoparallel:
Creating a general purpose event logger that will output event data to a CSV in a nicely formatted way.

Closes #515


jeamick:
Hello,

Does the goal is csv only or since you use polars there have different format like Json, parquet etc ?

kinrezC:
This feature has turned out to be substantially more difficult than expected.

in [6ef27dc](https://github.com/primitivefinance/arbiter/pull/516/commits/6ef27dc2fd9a942b08d6402ed30ee18569eb1100) the `EventCapture` is creating an independent tokio runtime on an independent system thread.  This will cause the results of the `EventCapture` to be inconsistent with the number of events that are emitted on the thread on which the environment is running.

There are two ways we can guarantee consistency between the `EventCapture` and `Environment`:
1. For every transaction, block the environment from making progress until the `EventCapture` has processed the event
2. Have the environment maintain a pointer to a value that tracks unique events, environment `Shutdown` state can only occur once the `EventCapture` has signaled to the environment that it has processed all of the events.

arbiter-core/src/tests/data_output.rs
```diff
@@ -0,0 +1,22 @@
+use super::*;
+use crate::data_collection::EventCapture;
+use futures_util::SinkExt;
+
+#[tokio::test]
+async fn data_capture() {
```
0xJepsen, line 6: Clean DX here

0xJepsen:
Can we close this in favor of #561 ?

PR #523, opened 2023-09-26
Alex/cheatcodes initial

### LINKED ISSUE #485: feat: use Foundry `Environment` based cheat codes
# Idea
In `arbiter-core` v0.5.0 we will have introduced a means to move block numbers and timestamps forward. However, we can do this (using `warp` and `roll`) and more by implementing the [environment cheatcodes](https://book.getfoundry.sh/cheatcodes/) provided by Foundry. There are a few others, at the very least, that will be helpful in managing our own environments and emulation. For instance `deal` will pass out raw ether to a specific address.

We would like to have more control over the database we are working with and give users more features from Foundry that they know and love.

### Importing Foundry EVM
First, let's note that all the relevant code for the cheatcodes lie in this [directory](https://github.com/foundry-rs/foundry/tree/5cfed63071ef1bb06f90190d2f678f546bfc971c/crates/evm/src/executor/inspector/cheatcodes). Stepping into the `mod.rs` we have the `Cheatcodes` struct which can contain a whole list of cheatcodes that can be applied. We can source the relevant crate by adding:
```toml
foundry-evm = { git = "https://github.com/foundry-rs/foundry.git" }
```

Note that this will require a bump in `tokio` via replacing our listed version with:
```toml
tokio = { version = "1.32.0", features = ["macros", "full"] }
```
It may be smart to pick a specific release of Foundry so that we do not have build issues moving forward.

Inside of `lib.rs` (most likely) we should then allow the cheatcodes module to be publicly usable by adding:
```rust
pub use foundry_evm::executor::inspector::cheatcodes::* as cheatcodes;
```
so that users using `arbiter-core` can simply have:
```rust
// now in their own project...
use arbiter_core::cheatcodes::*;
```
 
 ### Implementing Cheatcodes Inside `arbiter-core`
 
 #### The top level API
The cheatcodes should be accessible by a client that already has an interface to otherwise interact with the `Environment`. In v0.5.0 of `arbiter-core`, we have a user-facing function like this:
```rust
pub fn update_block(
        &self,
        block_number: impl Into<ethers::types::U256>,
        block_timestamp: impl Into<ethers::types::U256>,
    ) -> Result<ReceiptData, RevmMiddlewareError> {
        let block_number: ethers::types::U256 = block_number.into();
        let block_timestamp: ethers::types::U256 = block_timestamp.into();
        let provider = self.provider.as_ref();
        provider
            .instruction_sender
            .send(Instruction::BlockUpdate {
                block_number: block_number.into(),
                block_timestamp: block_timestamp.into(),
                outcome_sender: provider.outcome_sender.clone(),
            })
            .map_err(|e| RevmMiddlewareError::Send(e.to_string()))?;
        match provider.outcome_receiver.recv() {
            Ok(Ok(Outcome::BlockUpdateCompleted(receipt_data))) => Ok(receipt_data),
            _ => Err(RevmMiddlewareError::MissingData(
                "Block did not update Succesfully".to_string(),
            )),
        }
    }
```
Instead, we can replace this with a function that provides access to applying arbitrary Environment cheatcodes. In this case, a user can craft a cheatcode of their own by doing something along the lines of:
```rust
// inside of their own project
// pseudocode below
use arbiter_core::cheatcodes::Cheatcodes;
let cheatcode = Cheatcodes::new(something_goes_here);
client.apply_cheatcode(cheatcode);
```

#### Lower Level Implementation
We can bootstrap the work already done that applies the block update and send the `Instruction` over the channel to the `Environment` that the client is connected to. Inside of `arbiter-core` `environment.rs` module, we could make the change to the `Instruction` enum to be something like:
```rust
pub enum Instruction {
    Cheatcode(Cheatcode),
    Call {
        tx_env: TxEnv,
        outcome_sender: OutcomeSender,
    },
    Transaction {
        tx_env: TxEnv,
        outcome_sender: OutcomeSender,
    },
}
```

Then inside of the `Environment::run()` function, we would need to edit this match leg to reflect this somehow:
```rust
Instruction::Cheatcode(cheatcode) => {
                   // unbundle the cheatcode to see what it is doing
                   // check if it changes the block data, if so, make sure we are still in `UserControlled` variant.
                   // ~snip any other checks
                  cheatcode.apply_cheatcode(evm.db(), client.address(), 
                  }
```
Note that the above won't quite work, but it gets the point across. If you [look here](https://github.com/foundry-rs/foundry/blob/5cfed63071ef1bb06f90190d2f678f546bfc971c/crates/evm/src/executor/inspector/cheatcodes/mod.rs#L212), `Cheatcode` has a method that takes in a `EVMData` object which we can certainly construct from our `evm`'s DB. Most of it is changing the `evm.env` data. This should all be doable with (hopefully) minimal extra code written.

Finally, some kind of verification should be send as an output using this edited enum:
```rust
pub enum Outcome {
    CheatcodeCompleted(SomeDataToReturn),
    CallCompleted(ExecutionResult),
    TransactionCompleted(ExecutionResult, ReceiptData),
}
```

--- 

## Further notes

- I'm realizing now we might just want to `pub use ..:cheatcodes::env::*`.
- Notice now here we have a PUBLIC function [`apply()` here](https://github.com/foundry-rs/foundry/blob/5cfed63071ef1bb06f90190d2f678f546bfc971c/crates/evm/src/executor/inspector/cheatcodes/env.rs#L322)`

Alexangelj:
Closes #485 by creating the "Cheatcode" item of the Instruction enum and a "Cheatcodes" enum with the implemented cheatcodes. Adds a method `apply_cheatcode` to the client to send cheatcode instructions to the enviroment, very straightforward stuff!

#485 

**Changes**
- Removes Instruction::Deal
- Removes Outcome::DealCompleted
- Removes Middleware::deal()
- Adds Instruction::Cheatcode
- Adds Cheatcodes enum
- Adds Middleware::apply_cheatcode()
- Implements the Store, Load, and Deal cheatcodes in the environment instruction processing switch
- Adds a test for the store and load cheatcodes
- Edits the deal tests to use the new apply_cheatcode method


Alexangelj:
It's not quite ready - Load cheatcode doesn't work because apply_cheatcode method does not return a value, which some cheatcodes might. Will need to think about how to handle this, since the return values could be non-existent or a certain type.

0xJepsen:
> It's not quite ready - Load cheatcode doesn't work because apply_cheatcode method does not return a value, which some cheatcodes might. Will need to think about how to handle this, since the return values could be non-existent or a certain type.

Could we have the `apply_cheatcode` return an option? Or does it need to return something at since it sends results to the sender?

0xJepsen (commented):
Asked some questions

Alexangelj:
> > It's not quite ready - Load cheatcode doesn't work because apply_cheatcode method does not return a value, which some cheatcodes might. Will need to think about how to handle this, since the return values could be non-existent or a certain type.
> 
> Could we have the `apply_cheatcode` return an option? Or does it need to have it's signature changed at all since it sends result to the `outcomesender`?

Yeah I'm leaning towards not changing the signature and just using the outcome sender stuff to get the result - but trying to do that and running into some errors, so fixing that

Alexangelj:
Figured out the bug - since we are "catching" the results of applying a cheatcode, if we fail to catch one of the apply's we might get the return value of a different cheatcode, e.g.:

cheatcode::store -> dont catch
cheatcode::load -> catch

we catch the store result here, not the load result

0xJepsen:
> if we fail to catch one of the apply's we might get the return value of a different cheatcode, e.g.:



> Figured out the bug - since we are "catching" the results of applying a cheatcode, if we fail to catch one of the apply's we might get the return value of a different cheatcode, e.g.:
> 
> cheatcode::store -> dont catch cheatcode::load -> catch
> 
> we catch the store result here, not the load result
> 
> edit: it's not the best developer experience to remind everyone they need to catch all the results

Jus to clarify, are you saying that it's not a great devex to propagate the errors back with results? What do you mean by catch

Alexangelj:
> > if we fail to catch one of the apply's we might get the return value of a different cheatcode, e.g.:
> 
> > Figured out the bug - since we are "catching" the results of applying a cheatcode, if we fail to catch one of the apply's we might get the return value of a different cheatcode, e.g.:
> > cheatcode::store -> dont catch cheatcode::load -> catch
> > we catch the store result here, not the load result
> > edit: it's not the best developer experience to remind everyone they need to catch all the results
> 
> Jus to clarify, are you saying that it's not a great devex to propagate the errors back with results? What do you mean by catch

Yeah I mean if you don't catch it, and you try to catch something else later, you might catch the result you didnt catch earlier in the outcome sender recv

Alexangelj:
I fixed it with my latest commit - for some reason the submodules also got pushed?

0xJepsen:
> I fixed it with my latest commit - for some reason the submodules also got pushed?

Oh man yeah this is a pain, i think @kinrezC has a nice trick to remove them recursively so they don't get pushed? After that, and the merge conflicts it should be good to go

Alexangelj:
Alright, good to review + merge!

0xJepsen (changes requested):
Only thing i have a problem with is the gitsubmodules, can we get those removed?

Alexangelj:
> Only thing i have a problem with is the gitsubmodules, can we get those removed?

Removed!

PR #527, opened 2023-09-27
vulnerability corpus

### LINKED ISSUE #526: docs: Initial Corpus of issues found with arbiter
# Idea
The idea is, that as more vulnerabilities are found with Arbiter, we want to showcase them to illustrate the power of arbiter as an auditing tool. We want these to be really clean with instructions on how to reproduce the issue and extensive documentation of the attack vectors

This document can live in the root arbiter repo and be referenced by the readme. We should draw some inspiration from audit reports.

0xJepsen:
Closes #526 

Still need to be able to reproduce the bug, working on this here: https://github.com/primitivefinance/portfolio_simulations/pull/39


Alexangelj (approved):
Looks good - for all the new vulns we find we can just pr into this list? do you think thats the best way to track this for now?

0xJepsen:
> Looks good - for all the new vulns we find we can just pr into this list? do you think thats the best way to track this for now?

Yeah that's what i was thinking.

Alexangelj:
> > Looks good - for all the new vulns we find we can just pr into this list? do you think thats the best way to track this for now?
> 
> Yeah that's what i was thinking.

Sounds good - also the instructions to reproduce the bug are nice

Issue #484, opened 2023-09-08
feat: Nonce Manager


0xJepsen:
# Idea

We ought to be incrementing the nonces for all accounts as they submit transactions as they prevent replay attacks that we don't want to come up in discovering emergent behavior. 

## Solution

There is a none manager trait that is generic over some middleware [here](https://github.com/gakonst/ethers-rs/blob/master/ethers-middleware/src/nonce_manager.rs) I think this is the correct approach to move forward with. 

## Alternative approaches
We may also be able to make a naive nonce manager if we so desire.

Autoparallel:
@0xJepsen do you want to take this issue?

PR #534, opened 2023-09-28
Feat nonce middleware

0xJepsen:
Will be closing #484


arbiter-core/src/environment/mod.rs
```diff
@@ -556,6 +562,18 @@ impl Environment {
                                     )),
                                 }
                             }
+                            EnvironmentData::TransactionCount(address) => {
+                                let db = evm.db().unwrap();
+                                let recast_address = revm::primitives::Address::from(address);
+                                match db.accounts.get(&recast_address) {
```
Autoparallel, line 568:
```suggestion
                                match db.accounts.get(&address.into()) {
```

arbiter-core/src/environment/mod.rs
```diff
@@ -556,6 +562,18 @@ impl Environment {
...
+                                    Some(account) => {
+                                        Ok(Outcome::QueryReturn(account.info.nonce.to_string()))
```
Autoparallel, line 570: You could just set up its own return type so you don't have to stringify

arbiter-core/src/middleware/cast.rs
```diff
@@ -1,4 +1,4 @@
-///! Utility functions for casting between revm and ethers-rs types.
+/// ! Utility functions for casting between revm and ethers-rs types.
```
Autoparallel, line 1: I don't think this comment makes sense and I'm not sure what the `!` is for

arbiter-core/src/middleware/connections.rs
```diff
@@ -1,4 +1,4 @@
-///! Messengers/connections to the underlying EVM in the environment.
+/// ! Messengers/connections to the underlying EVM in the environment.
```
Autoparallel, line 1: I'm confused as to all the `/// !` comments. 

Module level doc comments are `//!` and other doc comments are usually just `///`

arbiter-core/src/tests/middleware_instructions.rs
```diff
@@ -300,6 +303,117 @@ async fn set_gas_price() {
...
+    let nonce_after_tx = client
+        .get_transaction_count(client.address(), Some(0.into()))
+        .await
+        .unwrap();
+    assert_eq!(nonce_after_tx.as_u64(), 1_u64);
+}
+
+#[tokio::test]
+async fn create_nonce_middleware() {
+    let (_manager, client) = startup_user_controlled().unwrap();
```
Autoparallel, line 325: All these spots you have `_manager` are not correct. The startup functions return `(Environment, Arc<RevmMiddleware>)`

arbiter-core/src/middleware/nonce_middleware.rs
```diff
@@ -0,0 +1,206 @@
...
+    address: Address,
+}
+
+impl<M> NonceManagerMiddleware<M>
+where
+    M: Middleware,
+{
+    /// Instantiates the nonce manager with a 0 nonce. The `address` should be
+    /// the address which you'll be sending transactions from
+    pub fn new(inner: M, address: Address) -> Self {
```
Autoparallel, line 37: Do we have any better ideas for setting up this API for creating a none managed middleware? 

@kinrezC do you have thoughts?

Autoparallel (changes requested):

arbiter-core/src/middleware/cast.rs
0xJepsen: I didn't write this comment, but it looks like `cargo fmt` hit it. I think it makes sense as a module comment doesn't it?

arbiter-core/src/environment/mod.rs
0xJepsen: So this gives some compile errors about needing additional type annotations. So it would need something like this
```rust
match db.accounts.get::<revm::primitives::Address>(&address.into())
```
Also I was following the convention from above that was used in the balance quarry. If we change it here and use the type annotations wouldn't it make sense to change everywhere?

arbiter-core/src/environment/mod.rs
0xJepsen: Why would we have a unique type for this one but not any of the others?

arbiter-core/src/middleware/connections.rs
0xJepsen: Same as above

arbiter-core/src/middleware/nonce_middleware.rs
0xJepsen: This is the way ether-rs creates theirs.

arbiter-core/src/environment/mod.rs
Autoparallel: I'm confused then. Why did you not use the `EnvironmentData` enum and the `Instruction::Query`? It seems weird to use a mix of the two. That's why I don't understand this.

arbiter-core/src/middleware/nonce_middleware.rs
Autoparallel: I'm not saying to change this method, I'm saying that giving the user an easy way to create a nonce managed middleware that is like an agent EOA could be a better devex. If this is the only entry point, then we might want to consider something additional.

arbiter-core/src/environment/mod.rs
Autoparallel: Sure, I was just suggesting something that is less LOC

arbiter-core/src/middleware/nonce_middleware.rs
0xJepsen: sure, but that seems like a different issue than the one that this feature close.

arbiter-core/src/environment/mod.rs
0xJepsen: I don't understand what you mean, every quarry does that. It's the exact same convention of all the other quarries no?

arbiter-core/src/environment/instruction.rs
```diff
@@ -149,6 +149,9 @@ pub(crate) enum EnvironmentData {
 
     /// The query is for the balance of an account given by the inner `Address`.
     Balance(ethers::types::Address),
+
+    /// The query is for the nonce of an account given by the inner `Address`.
+    TransactionCount(ethers::types::Address),
```
Autoparallel, line 154: @0xJepsen 

See here. You added a new instruction when there is already a `Instruction::Query`. Yet, you used the `Outcome::QueryReturn` which is just a bit confusing.

If you want to make a new instruction, I think it's best to make a new outcome. If you want to have a query outcome, then I think you should use a query instruction.

Autoparallel (changes requested):

arbiter-core/src/middleware/nonce_middleware.rs
Autoparallel: See my original comment. I just asked if there are better ideas for setting up the API.

arbiter-core/src/environment/instruction.rs
0xJepsen: I don't understand what you mean. The transaction count defined here is as described in it's comments right here
```
/// [`EnvironmentData`] is an enum used inside of the [`Instruction::Query`] to
/// specify what data should be returned to the user.
/// Currently this may be the block number, block timestamp, gas price, or
/// balance of an account.
```
Which is exactly how it's used and why i used it the way i did. It's exactly the same as the implementation for the `get_gas_price and ()` and `get_block_number()`?

Autoparallel (approved):

Autoparallel (approved):

PR #535, opened 2023-09-29
Feat bind for chosen contracts

### LINKED ISSUE #465: feat: add ability to generate bindings for chosen contracts
# Idea
Running `forge bind` may generate a lot of additional contracts that one does not want in their simulation crate. It would be helpful to provide a means via a `arbiter.toml` or even just an addition to `foundry.toml` that lists which contract bindings we want specifically. 

When running something like `forge bind` or `arbiter bind` there can be a massive amount of contract bindings generated when we may only need, say, 5 or so contracts.

## Solution
We can provide a typical bind interface and a config file of some sort where a user lists which specific bindings they want to keep. Upon running `bind` we can output files to a temporary location, grep for the contracts the user wants to keep, adjust the `mod.rs` file, then move this into the `bindings/` module directory in their simulation crate.

## Alternative approaches
We could propose changes to `forge bind` itself and attempt to get this changed at that level.

## Additional context
Bug report #439 is very related.

0xJepsen:
Closes #465 

This PR filters out bindings for any interface, test, or other ancillary contracts while ensuring the shared types are still generated. 

---

<details open="true"><summary>Generated summary (powered by <a href="https://app.graphite.dev">Graphite</a>)</summary>

> ## TL;DR
> This pull request updates the version of the "arbiter" package to 0.3.4 and adds a new dependency "tempfile" with version 3.8.0 to the dev-dependencies section in the Cargo.toml file. It also modifies the code to use only one codegen unit, collect a list of contracts from a specific directory, remove unnecessary contracts, and generate bindings for submodules.
> 
> ## What changed
> - Updated the version of the "arbiter" package to 0.3.4 in the Cargo.toml file.
> - Added a new dependency "tempfile" with version 3.8.0 to the dev-dependencies section in the Cargo.toml file.
> - Modified the code to use only one codegen unit.
> - Modified the code to collect a list of contracts from a specific directory.
> - Modified the code to remove unnecessary contracts.
> - Modified the code to generate bindings for submodules.
> 
> ## How to test
> 1. Execute the code snippet that checks if the output of a command is successful and prints the output if it is.
> 2. Execute the code snippet that removes unnecessary contracts and generates bindings for submodules.
> 3. Execute the code snippet that generates bindings for submodules in a given directory.
> 4. Execute the code snippet that sets the current directory to a given path and generates an output path based on the file name of the path.
> 5. Execute the code snippet that checks if the directory name is "src" or "contracts" and sets the target directory accordingly.
> 6. Execute the code snippet that checks if a filename does not start with 'I', converts the filename from camel case to snake case, and adds it to a list of contracts.
> 7. Execute the code snippet that removes any contracts that are not needed from a given directory.
> 8. Execute the code snippet that checks if certain contracts are needed and removes any contract bindings that are not needed.
> 9. Execute the code snippet that modifies a mod.rs file by removing certain lines based on specific conditions.
> 10. Execute the code snippet that converts a string from camel case to snake case.
> 11. Execute the code snippet that creates a temporary directory and nested directories within it.
> 12. Execute the code snippet that creates a temporary directory, creates files within the "src" directory, and collects a list of contracts from the directory.
> 13. Execute the code snippet that creates a temporary directory, creates files within the "src" directory, and updates the mod.rs file based on a list of contracts to keep.
> 14. Execute the code snippet that updates a mod file by removing certain contracts and adding others.
> 15. Check the content of the updated mod file to ensure that the removed contracts are no longer present and the added contracts are included.
> 
> ## Why make this change
> - Using only one codegen unit will slow down compilation but improve optimization.
> - Collecting a list of contracts from a specific directory allows for more flexibility in generating bindings.
> - Removing unnecessary contracts improves code cleanliness and reduces unnecessary overhead.
> - Generating bindings for submodules ensures that all necessary bindings are available for use.
</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ exclude = ["benches"]
 # Package configuration
 [package]
 name = "arbiter"
-version = "0.3.3"
+version = "0.3.4"
 edition = "2021"
 authors = ["Waylon Jepsen <waylonjepsen1@gmail.com>", "Colin Roberts <colin@autoparallel.xyz>"]
 description = "Allowing smart contract developers to do simulation driven development via an EVM emulator"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ thiserror = "1.0.40"
 # Dependencies for the test build and development
 [dev-dependencies]
 assert_cmd = "2.0.11"
+tempfile = "3.8.0"
 
 # Release profile
 [profile.release]
@@ -45,3 +46,5 @@ lto = true
 # The Rust compiler splits your crate into multiple codegen units to parallelize (and thus speed up) compilation but at the cost of optimization. 
 # This setting tells the compiler to use only one codegen unit, which will slow down compilation but improve optimization.
 codegen-units = 1
+
+
--- a/bin/bind.rs
+++ b/bin/bind.rs
@@ -34,7 +34,7 @@ pub(crate) fn forge_bind() -> std::io::Result<()> {
         .arg("--module")
         .arg("--overwrite")
         .output()?;
-    let mut contracts = collect_contract_list(Path::new("contracts"))?;
+    let project_contracts = collect_contract_list(Path::new("contracts"))?;
     if output.status.success() {
         let output_str = String::from_utf8_lossy(&output.stdout);
         println!("Command output: {}", output_str);
@@ -48,19 +48,20 @@ pub(crate) fn forge_bind() -> std::io::Result<()> {
         ));
     }
 
-    // Define path to `lib` directory
+    let src_binding_dir = Path::new("src/bindings");
+    remove_unneeded_contracts(src_binding_dir, project_contracts)?;
+
     let lib_dir = Path::new("lib");
-    let mut list2 = bindings_for_submodules(lib_dir)?;
-    contracts.append(&mut list2);
-    println!("Contract List: {:?}", contracts);
-    remove_unneeded_contracts(contracts)?;
+    let (output_path, sub_module_contracts) = bindings_for_submodules(lib_dir)?;
+    println!("submodule contracts: {:?}", sub_module_contracts);
+    remove_unneeded_contracts(Path::new(&output_path), sub_module_contracts)?;
 
     Ok(())
 }
 
-fn bindings_for_submodules(dir: &Path) -> io::Result<Vec<String>> {
+fn bindings_for_submodules(dir: &Path) -> io::Result<(String, Vec<String>)> {
     let mut contracts_to_generate = Vec::new(); // to keep track of contracts we're generating bindings for
-
+    let mut output_path = String::new();
     if dir.is_dir() {
         // Iterate through entries in the directory
         for entry in fs::read_dir(dir)? {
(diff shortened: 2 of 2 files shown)


Alexangelj (commented):
Nice work on this - I don't think I have enough context to really give review to this but some notes:
- structure to how this works is good, you specify the directory path and it uses all the contracts in there?
- We should rely on the foundry.toml's src, since that will be where most users specify what their contracts directory is, either src or contracts, or even something else.

Autoparallel (changes requested):
I don't really know how I would use this. Can you give some instructions with the commands so that if I do `arbiter bind --help` I can see how to use this?

An example in some doctests or something would also be nice.

0xJepsen:
> Nice work on this - I don't think I have enough context to really give review to this but some notes:
> 
> * structure to how this works is good, you specify the directory path and it uses all the contracts in there?
> * We should rely on the foundry.toml's src, since that will be where most users specify what their contracts directory is, either src or contracts, or even something else.

It does use the foundry toml for both the project contracts and the submodules

0xJepsen:
> I don't really know how I would use this. Can you give some instructions with the commands so that if I do `arbiter bind --help` I can see how to use this?
> 
> An example in some doctests or something would also be nice.

The utilization doesn't change at all, it just now only generates the bindings for the contracts you want to use not for anything like interfaces or libraries or stuff that was always bogging down our workflow.

Alexangelj:
> > I don't really know how I would use this. Can you give some instructions with the commands so that if I do `arbiter bind --help` I can see how to use this?
> > An example in some doctests or something would also be nice.
> 
> The utilization doesn't change at all, it just now only generates the bindings for the contracts you want to use not for anything like interfaces or libraries or stuff that was always bogging down our workflow.

How does the selection work? I'm assuming its just defaulting to the contracts that are in the src or contracts directory, and filtering out every other dependency contract

0xJepsen:
> > > I don't really know how I would use this. Can you give some instructions with the commands so that if I do `arbiter bind --help` I can see how to use this?
> > > An example in some doctests or something would also be nice.
> > 
> > 
> > The utilization doesn't change at all, it just now only generates the bindings for the contracts you want to use not for anything like interfaces or libraries or stuff that was always bogging down our workflow.
> 
> How does the selection work? I'm assuming its just defaulting to the contracts that are in the src or contracts directory, and filtering out every other dependency contract

`arbiter bind` will auto generate bindings for both contracts in the projects contract directory (Specified by the `foundry.toml` as the src, from `arbiter init`) and contracts for submodules (Also with a directory specified by the `foundry.toml` of the submod). 

The project contracts will be out to `src/bindings`
All submodule bindings will be out to a seperate module `src/{Submodule_Name}_bindings`

All ancillary and unnecessary contracts bindings (library, interface, and tests) will be removed and the changes will be reflected in the mod.rs for each binding module. This ensures that the binding modules are lean and not overwhelming or cluttered making them easier to parse and work with.

Autoparallel:
I think that's why I'm confused. From #465: 
> We can provide a typical bind interface and a config file of some sort where a user lists which specific bindings they want to keep. Upon running bind we can output files to a temporary location, grep for the contracts the user wants to keep, adjust the mod.rs file, then move this into the bindings/ module directory in their simulation crate.

Can I pick which contracts I'd like to keep by providing a list somewhere? This is what I wasn't sure about. I don't see how to do that. This is just removing libs, interfaces, and tests but not letting a user specify which bindings they want in particular.

0xJepsen:
> I think that's why I'm confused. From #465:
> 
> > We can provide a typical bind interface and a config file of some sort where a user lists which specific bindings they want to keep. Upon running bind we can output files to a temporary location, grep for the contracts the user wants to keep, adjust the mod.rs file, then move this into the bindings/ module directory in their simulation crate.
> 
> Can I pick which contracts I'd like to keep by providing a list somewhere? This is what I wasn't sure about. I don't see how to do that. This is just removing libs, interfaces, and tests but not letting a user specify which bindings they want in particular.

As specified by the issue description the problem was that the additional contracts were being generated. Now with only the needed bindings, does it not make sense to just give them all of them? since there are all lean and small and easy to manager?

PR #537, opened 2023-09-30
ft: fix tomls

### LINKED ISSUE #536: chore: Lock dependency versions
# Idea
so rather than something like
```toml
revm = "3.3.0"
```
we should lock the versions by prefacing the version with `=`
```toml
revm = { version = "=3.3.0" }
```
This ensures that no user experiences any ancillary breaking changes from upstream dependancies and we can be confident everything will always work and can be deliberate about adding support for core dependency major and minor changes

## Solution

comb through all the `.toml` files and fix the versions

0xethsign:
ft: bumped a lot of crates to latest version. initialised `blob_hashes` and `max_fee_per_blob_gas` with default values wherever necessary.

closes #536

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,19 +24,19 @@ path = "bin/main.rs"
 # Dependencies for the release build
 [dependencies]
 # Command line and config
-clap = { version = "4.3.0", features = ["derive"] }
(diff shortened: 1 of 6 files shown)


0xJepsen (approved):
This looks good to me! Thanks for taking care of this!

PR #540, opened 2023-10-01
:wrench: chore: update Solidity dependencies

### LINKED ISSUE #541: chore: fix `foundry.toml` remappings
The remappings in `foundry.toml` are deprecated. They use a syntax not used in `contracts/` and also include the no longer used Uniswap repos. These remappings should be corrected. Additionally, there is no `.gitmodules` tracking the Solidity dependencies.

Sabnock01:
Removes the now deprecated remappings for Uniswap in `foundry.toml`. 

Although it looks like forge-std isn't used anywhere either, I suppose there may be future need for it.

**Update**: Also pushed a commit to fix the `.gitmodules` which was currently missing and modified `foundry.toml` to use the correct import style to match `contracts/`.

Resolves #541


0xJepsen:
Shall we maybe close this in favor of #573 ?

Sabnock01:
> Shall we maybe close this in favor of #573 ?

That depends on how you plan to use the contracts. Do you want them to work out of the box? The two PR's don't really overlap. #573 is basically to just get rid of warnings / errors in the Solidity files. I'll leave to your discretion.

0xJepsen:
> > Shall we maybe close this in favor of #573 ?
> 
> That depends on how you plan to use the contracts. Do you want them to work out of the box? The two PR's don't really overlap. #573 is basically to just get rid of warnings / errors in the Solidity files. I'll leave to your discretion.

No we never use these contracts directly at all. We just export their bindings with the crate. We might not even ever really need the foundry.toml in this repo either.

Sabnock01:
> > > Shall we maybe close this in favor of #573 ?
> > 
> > That depends on how you plan to use the contracts. Do you want them to work out of the box? The two PR's don't really overlap. #573 is basically to just get rid of warnings / errors in the Solidity files. I'll leave to your discretion.
> 
> No we never use these contracts directly at all. We just export their bindings with the crate. We might not even ever really need the foundry.toml in this repo either. 

Fair enough. Will close.

PR #561, opened 2023-10-02
Event Logger

kinrezC:
Introduces `EventLogger`.  `EventLogger` has an `EventLoggerBuilder` which allows one to chain `EventLogger::builder().add(contract0.events()).add(contract1.events())` to listen for events on specified contracts.  When the `EventLogger` is `run().await` it will watch for events and write them to a csv.  

For #515


0xJepsen (approved):
This is so clean @kinrezC! Awesome work on this! You are officially the async rust chad.

arbiter-core/src/tests/data_output.rs
Autoparallel, line 1: My comments are mostly here.

Can we perhaps have these tests check that the files are actually created and have the correct data in them?

We don't need to take it super far, but I feel like these don't check all that much other than that this runs.

Autoparallel (changes requested):
How is the stop signal sent to the data collector? Do we need that ever? Can you explain why/why not?

Alexangelj (approved):
Looks excellent

PR #562, opened 2023-10-02
auto run + contructor return arc

### LINKED ISSUE #560: feat: Autostart the environment upon creation.
# Idea

Currently when building a simulation we have the end user start the environment after they create it 
```rust
    let mut manager = Manager::new();
    let _ = manager.add_environment(EnvironmentParameters {
        label: TEST_ENV_LABEL.to_owned(),
        block_settings: BlockSettings::UserControlled,
        gas_settings: GasSettings::UserControlled,
    });
    manager.start_environment(TEST_ENV_LABEL)?;
 ```
# Solution 
We should auto start the environment when we create it to simplify end user API

0xJepsen:
This PR Closes #560 

It also modifies the `RevmMiddleware` Constructor to return an `Arc<RevmMiddleware>` To improve the ease of use a bit since i don't believe there is a time in which we don't do this.

This does also have passing tests, clippy, and fmt

### DIFF
--- a/arbiter-core/Cargo.toml
+++ b/arbiter-core/Cargo.toml
@@ -17,8 +17,8 @@ revm = { version = "=3.4.0", features = ["serde"] }
 
 # Serialization
 bytes = { version = "=1.5.0" }
-serde = { version = "1.0.163", features = ["derive"] }
-serde_json = { version = "1.0.96" }
+serde = { version = "=1.0.163", features = ["derive"] }
+serde_json = { version = "=1.0.96" }
 
 # Concurrency/async
 tokio = { version = "=1.32.0", features = ["macros", "full"] }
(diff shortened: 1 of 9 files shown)


Autoparallel (changes requested):

Alexangelj (approved):
Is there any reason we didn't wrap it in Arc before?

Changes look good to me.

Also, maybe we can fix this: https://github.com/primitivefinance/arbiter/pull/530/commits/1e4988381f4050aef86fc610977990a3f630d1b3#diff-88c2b26ea7671c2bba8c63c1d8c8c6cefc45a3cef8e1cf9cea2610fefcbbb156R174

Those changes broke the branch w/ the new crate versions

Autoparallel:
> Is there any reason we didn't wrap it in Arc before?

Typically `new()` methods return `Self`. So it is a bit odd to `-> Arc<Self>` 

However, in this case there is no reason to do it otherwise.

Autoparallel (approved):
good cleanup

looking forward to writing less code for sims

PR #563, opened 2023-10-02
feat(arbiter-bin) forking state, disk i/o

### LINKED ISSUE #461: feat(arbiter-bin): add ability to fork chain state and store db
# Idea
Users **need** the ability to take the chain at a certain state, and fork it to be able to work with it in Arbiter. This is an essential, and highly requested feature. 

The problem we have is that all of the chain state we create is new. It would be nice to, for example, take a fork of Ethereum mainnet with the state of a few contracts at a certain block height. This would let people not only start up simulations more quickly, but let them work with historical and recent data.

## Solution
The solution is not exactly clear to me, but it is certainly possible to load the state of contracts into their respective addresses with `revm`. From there, we can use our `RevmMiddleware` to make interactions simple, e.g., create an instance of a contract via:
```rust
let contract = ContractStruct::new(contract_address_from_mainnet, revm_middleware_client);
```

## Alternative approaches
We can look into the methodology that Anvil uses to create forks, or perhaps even make Anvil the default blockchain environment for cases like this.

Autoparallel:
This PR provides a new command `fork` for the Arbiter binary that allows for a user to fork state of contracts/storage from a certain block from an Ethereum chain. Everything for the fork will be set up in a `.toml` file, and an example one `test_config.toml` is given.

Closes #461


Cargo.toml
```diff
@@ -24,20 +24,32 @@ path = "bin/main.rs"
 # Dependencies for the release build
 [dependencies]
 # Command line and config
-clap = { version = "=4.3.0", features = ["derive"] }
-serde = { version = "=1.0.163", features = ["derive"] }
-toml = { version = "=0.8.0" }
+clap = { version = "4.3.0", features = ["derive"] }
+toml = "0.8.0"
```
0xJepsen, line 28: We should fix all these versions probably no?

Cargo.toml
```diff
@@ -24,20 +24,32 @@ path = "bin/main.rs"
...
 thiserror = { version = "=1.0.47" }
 
+# Forking
+ethers = "2.0.10"
+revm = { version = "3.3.0", features = ["serde", "ethersdb"] }
+
+# General serialization
+serde = { version = "1.0.163", features = ["derive"] }
+serde_json = "1.0.107"
+strum = "=0.25.0" # MIGHT NOT NEED
```
0xJepsen, line 44: I see we are using `strum_macros`, but didn't see `strum` anywhere

bin/main.rs
```diff
@@ -22,8 +22,13 @@ use std::error::Error;
 use clap::{command, CommandFactory, Parser, Subcommand};
 use thiserror::Error;
 
+use crate::fork::{digest::ForkConfig, ForkedDB};
```
0xJepsen, line 25: don't think we use ForkedDB anywhere

bin/fork/mod.rs
```diff
@@ -0,0 +1,101 @@
+use std::str::FromStr;
```
0xJepsen, line 1: unused import

bin/fork/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+use std::{
+    collections::HashMap,
+    fs::{self, File},
```
0xJepsen, line 4: File is unused

0xJepsen (changes requested):
I noticed sometime things don't compile with this error:
```
error[E0599]: no method named `into_environment_parameters` found for struct `EnvironmentBuilder` in the current scope
   --> /Users/shufflebottomhogwood/.cargo/git/checkouts/arbiter-e286ca1df4b4c9d0/1f58eee/arbiter-core/src/environment/builder.rs:131:31
    |
52  | pub struct EnvironmentBuilder {
    | ----------------------------- method `into_environment_parameters` not found for this struct
...
131 |         let parameters = self.into_environment_parameters();
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `EnvironmentBuilder`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `arbiter-core` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
```

Alexangelj (approved):
Holy **** *** ********* ***

******* amazing

Looks excellent to me

0xJepsen (approved):
Hell yeah brother

PR #583, opened 2023-10-03
Arbiter/version

### LINKED ISSUE #471: feat: refactor `EnvironmentParameters` to include label

### LINKED ISSUE #484: feat: Nonce Manager

### LINKED ISSUE #485: feat: use Foundry `Environment` based cheat codes

### LINKED ISSUE #511: feat: optional block time
## Idea

Make block time optional when instantiating an Environment/BlockSettings. Sometimes want to abstract time away and just work in a block. Mandatory block time makes that harder.

## Solution

- make block_time field optional
- interpret a None value to disable time-based block progression

## Alternative approaches

maybe instead of an Option, use a custom enum so we can differentiate modes like "mine on tx," "mine per time," or "never mine automatically"

### LINKED ISSUE #513: feat: remove `Manager` and use builder pattern for `Environment`
# Idea

Mentioned by @kinrezC earlier in team chat. I think the UX for this could be better by using something like

```rust
let env = Environment::new().block_settings(some_settings).gas_settings(other_settings).label("label".into()).start()?;
```

## Solution

Remove constructor parameters from the environment constructor and build the setters. 

## Things to consider

What should the behavior be when someone attempts to start the `env` when not gas / block / label, are not set?

edit by @kinrezC 
Tasks included in this issue:
1. Remove `manager`, sims will now be created and run via the creation of an `Environment`. 
2. Add defaults for `Environment::{BlockSettings, GasSettings}` to be `UserControlled`
3. Make `Environment::label` optional for tagging on concurrent sims
4. Refactor `Environment` to use builder pattern for modifying settings and adding a label

### LINKED ISSUE #515: feat: general purpose event logger

### LINKED ISSUE #524: chore(arbiter-core): refactor `environment.rs` module
# Idea
This single file has gotten far too large and it should be broken apart into logical components.

## Solution
Consider breaking into:
- `mod.rs`
- `tests.rs`
- `errors.rs`
- `instruction.rs`
- `builder.rs`

### LINKED ISSUE #528: refactor(arbiter-core): `middleware.rs` module and bug prevention
# Idea
The `middleware.rs` module has also gotten quite large. We can:
- Move errors to an `errors.rs` submodule.
- Have the main struct, struct impl, and trait impl in a `mod.rs`.
- Have another module for the ancillary objects?

## Bug prevention
- By setting `type Inner = Self;` we are creating place where we can hit a stack overflow as the default impls of middleware call `self.inner()`. We should instead have `type Inner = Provider<Connection>` as this is truly the next middleware stage and it is also the bottom middleware of the onion.

0xJepsen:
# Version 0.6.0

This will ship a collection of new features for 0.6.0

Closes #515 
Closes #513 
Closes #484
Closes #511 
Closes #485
Closes #471 
Closes #524
Closes #528


0xJepsen:
So two tests fail on occasion, not every time. These are the ones i have had fail after running the tests multiple times. I think for each one we should put them in a for loop and debug / make sure they are always passing. 

- [ ] `test tests::data_output::data_capture_output_validation ... FAILED`
- [ ] `test fork::tests::read_in ... FAILED`
- [ ] `test fork::tests::write_out ... FAILED`

PR #584, opened 2023-10-04
docs: setup and cli installation details

### LINKED ISSUE #557: :memo: Docs: (subsection)  CLI  installing with cargo and starting with `arbiter init` using `arbiter bind` submodules etc
This should add sufficient documentation to the installation section of the #538. It should cover installing arbiter with cargo, using `arbiter init` and `arbiter bind`, as well as using submodules with `forge install`

ruvaag:
* Adds docs to setup and install Arbiter CLI, building from source
* README typo fix

Resolves #557

### DIFF
--- a/docs/src/getting_started/cargo_install.md
+++ b/docs/src/getting_started/cargo_install.md
@@ -1,1 +1,43 @@
-# CLI installation with cargo
+# CLI Installation with Cargo
+
+## Prerequisites
+The best way to build Arbiter today is from source. 
+
+Install Git, if you haven't already. There are many online guides on how to install Git on different devices, including one from [Github](https://github.com/git-guides/install-git).
+
+Before installing Arbiter, ensure that you have Rust installed. You can install and verify your Rust installation from the [official website](https://www.rust-lang.org/tools/install).
+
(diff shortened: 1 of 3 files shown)


docs/src/getting_started/cargo_install.md
```diff
@@ -1 +1,43 @@
-# CLI installation with cargo
+# CLI Installation with Cargo
+
+## Prerequisites
+The best way to build Arbiter today is from source.
```
0xJepsen, line 4: I would actually say that it's much easier to install from crates with `cargo install arbiter` than from the source. I think it's good to include instructions for both but i crates is easier because you don't have to build from src.

0xJepsen (changes requested):
Looks really great! I think we are missing instructions on installing with cargo

docs/src/getting_started/cargo_install.md
ruvaag: good point. will rework this part and probably just nix this particular line

0xJepsen (approved):
Looks good!

0xJepsen:
Paging @Autoparallel or @kinrezC for a review here

Autoparallel (approved):
Awesome work! I love to see the docs come along :)

PR #592, opened 2023-10-05
feat: add `contracts` feature flag to `arbiter-core`

### LINKED ISSUE #573: feat: add feature flag to `arbiter-core/` crate for `contracts/`
# Idea
The files in `arbiter-core` are currently not being used anywhere in the project. We want to allow users the option to include them with the `arbiter-core` crate when using it as a dependency.

## Solution
A feature flag should be added to the [`arbiter-core`](https://github.com/primitivefinance/arbiter/tree/main/arbiter-core) crate to optionally include the files in `contracts/`.

Sabnock01:
Moves `contracts/` into `arbiter-core` and adds a feature flag to the crate to include the contracts. 

Resolves #573


Autoparallel (changes requested):
In the `arbiter-core` lib file, we need to annotate the `bindings` module like so:

```rust
#[cfg(feature = "contracts")]
pub mod bindings; 
pub mod data_collection;
pub mod environment;
pub mod math;
pub mod middleware;
#[cfg(test)]
mod tests;
```

This may interfere with some tests, so we should be careful of that.

Externally, we should find that only if we have:
```toml
arbiter-core = { git = "<your github link>", branch = "<your branch name>", features = "contracts" }
```
are we able to
```rust
use arbiter_core::bindings::*;
```

Alexangelj (approved):
I like these changes - this is how it should be done.

I know that the tests rely on the bindings of these contracts @Autoparallel , I think keeping some static bindings in the test folder is how we go about keeping the tests in without relying on the generated bindings from these contracts

Autoparallel (approved):

Autoparallel:
Nice work!

Sabnock01:
Feel free to squash

PR #593, opened 2023-10-05
Feat/alt event capture

kinrezC:
- Improved version of the event capture
- `.add` now accepts events from arbitrary contracts rather than enforcing that all events are from a `ContractInstance` of the same type
- Files are tracked in a BTreeMap to avoid re-opening when new events are received in the to the stream


Autoparallel (approved):
Looks good! Glad we got this going!

0xJepsen (approved):
looking good guys :)

PR #595, opened 2023-10-05
feat: bump revm to v3.5.0

### LINKED ISSUE #591: Chore: bump minor versions on dependancies
We want to get some downstream minor patches propagated into the repo the most exiting of which is the allow primitives from revm here [#revm](https://github.com/primitivefinance/arbiter/pull/589)

We may want to do this in the feature branch in case it causes a breaking change, but if it doesn't i am open to merging it into main as a patch.

ruvaag:
* Bump all dep minor versions: no changes needed, everything except revm was up to date
* Bump revm to 3.5.0

P.S - note uses of `to_be_bytes()` which could be a source of errors due to possible endianness incompatibility between alloy and ethers primitives

Resolves #591


Cargo.toml
```diff
@@ -26,13 +26,13 @@ path = "bin/main.rs"
 arbiter-core = { path = "arbiter-core" }
 
 # Command line and config
-clap = { version = "=4.3.0", features = ["derive"] }
+clap = { version = "=4.4.6", features = ["derive"] }
```
0xJepsen, line 29: Is this needed for the revm primitives?.

0xJepsen (approved):
im also down to merge this in, nice work!

PR #602, opened 2023-10-10
Patch: `toggle_written_columns` for multiple files

kinrezC:
Previous version had `toggle_written_columns` as a global bool for the handle.  This PR adds a new BTreeMap that tracks whether or not the headers are written for each file rather than for each handle.


0xJepsen (approved):
lgtm!

PR #607, opened 2023-10-12
Test workflow on macos-latest

kinrezC:
- Runs tests on 3 core macos-latest deployment rather than 2 core ubuntu-latest, see [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners)

edit: running the tests on macos makes data_capture pass

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     name: test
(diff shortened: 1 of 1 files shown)


kinrezC:
Looks like the example_fork got lost in translation at some point?

0xJepsen (approved):
im down to merge this

PR #620, opened 2023-10-16
Feature/forking eoa

### LINKED ISSUE #619: feat(arbiter): Add in ability to fork EOA storage, not just contract.
# Idea
At the moment `arbiter fork` only forks contracts and not EOA's. This is an oversight on my behalf. We should add in another field to the  `ForkConfig` that accepts optional `ExternallyOwnedAccounts`. These should be (hopefully) simpler than contracts, and follow most of the same boilerplate. We shouldn't need any artifacts for these as well!

Yes, in using this in trying some examples, we came across the inability to do this.

## Solution
Let's change `ForkConfig` to this:
```rust
pub(crate) struct ForkConfig {
    output_directory: Option<String>,
    output_filename: Option<String>,
    provider: String,
    block_number: u64,
    #[serde(rename = "contracts")]
    contracts_meta: HashMap<String, ContractMetadata>,
    externally_owned_accounts: HashMap<String, Address>,
}
```
Now in a `.toml` we can have:
```toml
[externally_owned_accounts.my_account]
address = "0x..."
```

There will be more work to then fetch the relevant storage for this account. I am not sure exactly how hard that will be! Hopefully not hard.

0xJepsen:
Add support to the fork CLI to be able to fork externally owned accounts

Closes #619 

---

<details open="true"><summary>Generated summary (powered by <a href="https://app.graphite.dev">Graphite</a>)</summary>

> ## TL;DR
> This pull request introduces the ability to fetch and store account information for externally owned accounts (EOAs) in the `ForkConfig` struct. It also includes minor changes to the test file and the example configuration file.
> 
> ## What changed
> - Added a new field `externally_owned_accounts` to the `ForkConfig` struct in `mod.rs`.
> - Added logic to fetch and store account information for each EOA in the `externally_owned_accounts` vector.
> - Removed the line in `tests.rs` that deletes the test file after reading it in.
> - Added an example EOA to the `weth_config.toml` file.
> 
> ## How to test
> 1. Pull the changes from this PR.
> 2. Run the tests in `tests.rs` to ensure they pass.
> 3. Use the updated `weth_config.toml` file to test the new functionality. The output should include account information for the specified EOA.
> 
> ## Why make this change
> This change allows users to fetch and store account information for EOAs, which can be useful for various purposes such as testing or data analysis. The removal of the file deletion line in the test file ensures that the test file is not deleted after each test run, making it easier to rerun tests. The changes to the example configuration file provide a practical example of how to use the new functionality.
</details>


example_fork/weth_config.toml
```diff
@@ -4,6 +4,9 @@ output_filename = "test.json"
 provider = "https://eth.llamarpc.com"
 block_number = 18228556
 
+# EOAs
+externally_owned_accounts = ["0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"] # vitalik.eth as of 10/2/2023
```
Autoparallel, line 8: I think this should be a hashmap so we can reference by string the eoas

Autoparallel (changes requested):

0xJepsen:
The requested cahnges have been resolved and logic has been added to the test to ensure the data is being fetched from the chain, written to storage, and read into revm correctly. The one thing i realized and am not sure how to work with is that we cannot make a wallet object to a particular address without the private key. From what i know the address is a has of the public key which is generated from the private key by multiplying a generator point in the elliptic curve field. by some pseudorandom number. I was looking at the [ethers-signers](https://github.com/gakonst/ethers-rs/blob/master/ethers-signers/src/wallet/mod.rs) to see if there were any mock wallets that bypassed this in tests but didn't find anything. We kinda want to just gas light revm into verifying the signature is right. Maybe we want to change wallet struct attribute of the provide to out own for this?

Autoparallel:
> The requested cahnges have been resolved and logic has been added to the test to ensure the data is being fetched from the chain, written to storage, and read into revm correctly. The one thing i realized and am not sure how to work with is that we cannot make a wallet object to a particular address without the private key. From what i know the address is a has of the public key which is generated from the private key by multiplying a generator point in the elliptic curve field. by some pseudorandom number. I was looking at the [ethers-signers](https://github.com/gakonst/ethers-rs/blob/master/ethers-signers/src/wallet/mod.rs) to see if there were any mock wallets that bypassed this in tests but didn't find anything. We kinda want to just gas light revm into verifying the signature is right. Maybe we want to change wallet struct attribute of the provide to out own for this?

This is true. We will need a new way to create a `RevmMiddleware`. We can just do something like
```rust
fn new_from_address(env: Environment, address: Address, label: Option<&str>) -> {
...
}
```

Nothing crazy.

Autoparallel (approved):

PR #626, opened 2023-10-20
Eoa forking

### LINKED ISSUE #624: feat(arbiter-core): create middleware from forked EOA
# Idea
We can now fork EOAs into Arbiter environments, but we have no easy way to give them middleware. So we need to do that.

## Solution
We can make a new function for `RevmMiddleware` like so:
```rust
impl RevmMiddleware {
  fn new_from_address(env: Environment, address: Address, label: Option<&str>) ->  Self {
      ...
  }
// ~ snip
```
The method body can follow a lot of the `RevmMiddleware::new()` we already have. The only difference is we won't create a wallet with a private key. Perhaps this means we should redesign some of the structure here a bit making wallets optional and addresses required. 

## Alternative approaches
We could create a whole different middleware for them too. I'm open to other suggestions.

0xJepsen:
Closes #624 

The approach i went with here was change the wallet property of the middleware to be an enum of either wallet or forked.

### DIFF
--- a/arbiter-core/Cargo.toml
+++ b/arbiter-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "arbiter-core"
-version = "0.6.2"
+version = "0.6.3"
 edition = "2021"
 authors = ["Waylon Jepsen <waylonjepsen1@gmail.com>", "Colin Roberts <colin@autoparallel.xyz>"]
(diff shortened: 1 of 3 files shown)


Autoparallel (approved):
Yeah this was an excellent way to do it. Nice work!

PR #630, opened 2023-10-20
Bind include interfaces

### LINKED ISSUE #628: Bind should include interface on default and then give option to exclude them
If we define a single interface for all our strategies we want to have the interface contracts included in the whitelist in arbiter bind. Let's include them by default.

0xJepsen:
closes #628 

This add a bool field to the toml table the we can set to want interfaces. By default it is turned off.


Alexangelj (approved):
looks great, config nice

PR #634, opened 2023-10-22
fix(bind): output directory outputs to src/bindings

Alexangelj:
With the arbiter config "bindings_workspace", it's taking the workspace as the root directory and adding "src/bindings", then at the end of the function it adds another "/bindings", so output for the workspace "simulation" is "simulation/src/bindings/bindings". 

I remove the first addition of bindings so we rely on the join at the end of the function.

image


0xJepsen:
Clean, yeah i was noticing this. Thanks for shipping this <3

0xJepsen (approved):

PR #660, opened 2023-10-27
fix test and change output directory

### LINKED ISSUE #657: Patch: Simout
Change the output directory of the event logging to be something other than `out/`

0xJepsen:
Closes #657 and fixes tests

- [x] updates CI to not use Macos (expensive)
- [x] Removes book deploy from CI


.github/workflows/book.yaml
```diff
@@ -109,24 +109,4 @@ jobs:
...
-      pages: write
-      id-token: write
-
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
-
-    steps:
-      - name: Deploy to GitHub Pages
-        id: deployment
```
Autoparallel, line 131: What all is being removed here?

Autoparallel (approved):

.github/workflows/book.yaml
0xJepsen: This is removing the deploy job for the book that we never run because our mdbook isn't deployed to github pages (yet)

PR #665, opened 2023-10-30
chore: toml fix

### LINKED ISSUE #646: Move Arbiter Config out of cargo.toml and into foundry.toml

0xJepsen:
Closes #646 

This also takes the fork tests out of a parallel loop so they will take load off the CI and fail less frequently


0xJepsen:
So this throws a foundry error if we put it there. Not sure where we should put this

0xJepsen:
Maybe move back to the arbiter.toml and we can put more higher level configuration there

bin/bind/test.rs
```diff
@@ -0,0 +1,108 @@
+#[allow(unused_imports)]
```
0xJepsen, line 1: Rename this file tests from test

PR #675, opened 2023-11-06
feat(arbiter-core): event logger metadata

### LINKED ISSUE #674: feat(arbiter-core): add metadata for the `EventLogger`
# Idea
When running simulations there will be some kind of metadata used to identify which file corresponds to which configuration. Providing a way to have this piped into the output file directly would be helpful!

## Solution
Add a `metadata` field to the `EventLogger` struct. The field need only be `Serialize` and can be generated along with the event content upon receiving the stop signal.

Autoparallel:
Adds a metadata field to the event logger and appends it to the JSON file output.

Closes #674


kinrezC (approved):

Alexangelj (approved):

0xJepsen (approved):
this looks good to me

PR #693, opened 2023-11-09
refactor: add additional logging within `arbiter-core`

### LINKED ISSUE #684: Add more logs to arbiter core
it would be nice for us to have more information from our tracing and logs or at least optionality for more verbosity.

### LINKED ISSUE #686: refactor(arbiter-core): remove re-export of `RustQuant`
# Idea
We don't need to re-export the modules of `RustQuant` that we're using. This just adds a bit of bloat and the end user can use it if they wish (as we do in our own in-house project). 

## Solution
We may need `RustQuant` for the stochastics we use in `arbiter-core` but we also may not (I will check).

Autoparallel:
Added more logging to the `arbiter-core` crate to make working with the crate easier. Also, removed the re-export of RustQuant as it isn't really necessary.

Closes #684 and closes #686


0xJepsen (approved):
I'm down to merge this if we resolve conflicts and get fmt to pass

PR #702, opened 2023-11-13
break event collector loop

kinrezC:
- Adds missing `break`s to break out of the event collector while loop when data is successfully written to file
- Adds logic for broadcasting a shutdown signal from `data_collection.rs` threads to the main environment thread to guarantee that output data is written to completion


0xJepsen (approved):
good kush

PR #711, opened 2023-11-15
feat(access-cheatcode): fetches the DbAccount of an address

Alexangelj:
# Changelog
- Adds the Access Cheatcode, which fetches and returns the `DbAccount` for a given address.

### DIFF
--- a/arbiter-core/src/environment/cheatcodes.rs
+++ b/arbiter-core/src/environment/cheatcodes.rs
@@ -3,6 +3,8 @@
 
 #![warn(missing_docs)]
 
+use revm_primitives::{AccountInfo, HashMap, U256};
+
 /// Cheatcodes are a direct way to access the underlying [`EVM`] environment and
 /// database.
 #[derive(Clone, Debug, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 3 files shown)


arbiter-core/src/environment/cheatcodes.rs
```diff
@@ -36,6 +38,28 @@ pub enum Cheatcodes {
...
+    /// Fetches the `DbAccount` account at the given address.
+    Access {
+        /// The address of the account to fetch.
+        address: ethers::types::Address,
+    },
+}
+
+/// Wrapper around [`AccountState`] that can be serialized and deserialized.
+#[derive(Debug, Clone, Default, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+pub enum AccountStateSerializable {
```
Autoparallel, line 50: Consider PRing this change into `revm`? https://github.com/bluealloy/revm/blob/310cabaa0a1306687dd4a8218a74c27fb338bbc8/crates/revm/src/db/in_memory_db.rs#L333

Autoparallel: Actually, I don't quite see why we need the serde traits now as I don't think they're used anywhere. Do you have plans to use them elsewhere?

PR #712, opened 2023-11-16
working on json output

### LINKED ISSUE #704: Migrate the json file writer in the event logger to use the polars file writer

0xJepsen:
I think a good way to go about this is to write a test that ensures that the ouput.json is the same as it was with the prior method. I can commit the two files here and maybe we go from there. 

Will close #704


Autoparallel:
Can I close this as stale or do you want to leave it open?

Issue #606, opened 2023-10-12
Feat: Change cheat-code types from ethers types to alloy types


0xJepsen:
# Idea
Currently in some of our cheatcodes for example in `Deal` we have the types specified as ethers types. It would be nice if we only exposed the end user to alloy types.

PR #725, opened 2023-11-27
feat: Change types from ethers types to alloy types in cheatcodes

quidproquoo:
This is a draft PR for the issue #606 

It seems like `BlockId` enum is no longer supported in alloy types. https://github.com/primitivefinance/arbiter/blob/c6ea6ab526afc1d9e349a501d2213d2b2d34115f/arbiter-core/src/environment/cheatcodes.rs#L28C9-L28C9

So what should be the approach here? Should I create a enum that wrappes both `BlockHash` and `BlockNumber` or do something else for that?

### DIFF
--- a/arbiter-core/src/environment/cheatcodes.rs
+++ b/arbiter-core/src/environment/cheatcodes.rs
@@ -3,7 +3,7 @@
 
 #![warn(missing_docs)]
 
-use revm_primitives::{AccountInfo, HashMap, U256};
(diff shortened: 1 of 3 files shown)


0xJepsen (approved):
lgtm

Autoparallel:
lgtm too.

Tests failing is not your problem. That's ours. I'll merge this now. 

Good work @quidproquoo

PR #736, opened 2023-12-05
bugfix: revert alloy and fix tests

Autoparallel:
We had found that the changes to Alloy primitives caused a handful of breaks in `arbiter-core`. These could not be resolved easily, so they have been reverted for now. This was found simultaneously as we had integration tests hidden behind a missing feature flag. Also also, the `--all` flag was deprecated around the same time in favor of `--workspace` for `cargo test` which caused us grief too. 

Apologies to @quidproquoo for reverting these changes. This is on us.

I have also `#[ignore]`d three tests that have to do with forking. Features here haven't changed, and these tests continuously eat up a lot of time and often break for reasons unknown.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -35,7 +35,7 @@ jobs:
           components: clippy
 
       - name: cargo clippy
-        run: cargo clippy --all --all-features -- -D warnings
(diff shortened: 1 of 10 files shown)


0xJepsen (approved):
lgtm glad you ignored the forking tests.

PR #744, opened 2023-12-13
feat(arbiter-core): db and coprocessor backend

Autoparallel:
Starting to expand access to the underlying EVM DB used in an `Environment`. This can be shared for parallel calls that do not mutate state, for snapshotting the db to pause and reload sims, and for outputting to disk.

TODO: 
- [x] Implement `serde` traits for `ArbiterDB`
- [x] `Environment::stop()` returns the `ArbiterDB`
- [x] Provide functionality to read/write from/to disk.
- [ ] **(NOT PLANNED FOR THIS PR)** Pipeline calls into a coprocessor to free the main tx thread (spawned in `Environment::run()`) from having to sequence non-mutating calls. This allows for parallelization.


Autoparallel:
Q: What if we just added a return of the `ArbiterDB` when we call `Environment::stop()`?

arbiter-core/src/environment/mod.rs
```diff
@@ -475,7 +482,7 @@ impl Environment {
                         // Set the tx_env and prepare to process it
                         evm.env.tx = tx_env;
 
-                        let result = evm.transact()?.result;
+                        let result = evm.transact_ref()?.result;
```
Alexangelj, line 485: Little confused on this, could you explain what happens here?

Alexangelj (dismissed):
looks very easy to use thank you!

arbiter-core/src/environment/mod.rs
Autoparallel: Yes. There are 3 traits for the DB's in Revm: `Database`, `DatabaseCommit`, and  `DatabaseRef`.

All the methods for both `Database` and `DatabaseCommit` take in the db as exclusive reference`&mut` whereas the methods in `DatabaseRef` take in only as ref `&`.  The reason why we changed from `evm.transact()` to `evm.transact_ref()` is because  the implementation I used for `ArbiterDB` for these traits uses `self.0.write()` for the `&mut` and `self.0.read()` for `&` for the `DatabaseRef` version. This is done so that if the environment is currently handling a read-only instruction here, then the `RwLock` is not blocking other read-only instructions from being used on shared state.

0xJepsen (approved):
lgtm

PR #746, opened 2023-12-13
feat(arbiter-engine): Run method for agents and messager echo example

### LINKED ISSUE #745: feat(arbiter-engine): create `Agent::run()` and an example

Autoparallel:
This PR adds the ability to have a world ran via `World::run()`. To show how this works, see `arbiter_engine::examples::tests`, specifically the `base_simulation()` function. This will run a simulation that loads an agent's engine with a messager as both a collector and executor which allows for this agent to receive external stimulus and communicate within this layer (potentially to other agents). Externally, "Start" is sent to the agent which kicks off a loop where the agent repeatedly echoes "Hello, world!" every second. This does not exit gracefully yet, use `ctrl + c`. It is a POC.

Closes #745

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -241,8 +241,6 @@ name = "arbiter-bindings"
 version = "0.1.1"
 dependencies = [
  "ethers",
- "revm",
- "revm-primitives",
  "serde",
 ]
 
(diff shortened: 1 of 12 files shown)


0xJepsen (approved):
Really nice

PR #749, opened 2023-12-15
enhance(arbiter-engine): agent behavior abstraction

### LINKED ISSUE #751: feat(arbiter-engine): example multi-agent simulation

### LINKED ISSUE #752: feat(arbiter-engine): allow multiple `Behavior`s for `Agent`

Autoparallel:
This PR brings in features that allow an `Agent` to have multiple `Behavior`s if that is desired. We demonstrate a multi-agent simulation -- to see it run:
```bash
cargo test --package arbiter-engine --lib --all-features -- examples::token_minter::token_minter_simulation --exact --nocapture --ignored
```
It will not stop on its own, but it should read out logs that show:
1. An `ArbiterToken` named "Arbiter Token" is deployed.
2. A "requester" asks the "token_admin" for the address of "Arbiter Token".
3. The "token_admin" replies back to this query with the data of "Arbiter Token" that includes the address.

- [x] Closes #752
- [x] Closes #751


arbiter-engine/src/examples/token_minter.rs
```diff
@@ -0,0 +1,315 @@
...
+    pub mint_amount: u64,
+}
+
+#[async_trait::async_trait]
+impl Strategy<Message, MessageOrTx> for TokenAdmin {
+    #[tracing::instrument(skip(self), fields(id = %self.id))]
+    async fn sync_state(&mut self) -> Result<()> {
+        debug!("Syncing state for `TokenAdmin`.");
+        for token_data in self.token_data.values_mut() {
+            let token = ArbiterToken::deploy(
```
0xJepsen, line 85: Nice i like this as the sync state.

arbiter-engine/src/examples/token_minter.rs
```diff
@@ -0,0 +1,315 @@
...
+    #[tracing::instrument(skip(self, event), fields(id = %self.id))]
+    async fn process_event(&mut self, event: Message) -> Vec<MessageOrTx> {
+        trace!("Processing event for `TokenAdmin` {:?}.", event);
+        if self.tokens.is_none() {
+            error!("There were no tokens to deploy! You must add tokens to the token admin before running the simulation.");
+        }
+        if event.to == self.id {
+            let query: TokenAdminQuery = serde_json::from_str(&event.data).unwrap();
+            trace!("Got query: {:?}", query);
+            match query {
```
0xJepsen, line 115: also thing that it's a good pattern to match on message type. Do you think in the future we could abstract some of the logic into helpers with descriptive to make it intuitive? or do you think thats not really important?

arbiter-engine/src/examples/token_minter.rs
```diff
@@ -0,0 +1,315 @@
...
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn token_minter_simulation() {
+    // TODO: Test outline, requester requests 1, 2, 3 tokens, and the test stops
+    // when the balance is checked to be 6.
+    std::env::set_var("RUST_LOG", "trace");
+    tracing_subscriber::fmt::init();
+
+    let environment = EnvironmentBuilder::new().build();
+    let connection = Connection::from(&environment);
+    let provider = Provider::new(connection);
```
0xJepsen, line 260: What do you think about just passing a reference of the environment to the world constructor? Not for any particular reason other than making the sim api lean. Or maybe we could even have the environment builder return a world?

0xJepsen: I guess that doesn't make much sense because the clients need to be constructed around the environment, nvm.

0xJepsen (approved):
lgtm

arbiter-engine/src/examples/token_minter.rs
Autoparallel: Other way to do it is to break it into multiple behaviors, each being simple and dealing with one type of event. I think this is also a valid pattern, however I did it this way because I was having to debug multi-agent stuff. I plan to revisit this example and display two ways to do it to see which UX seems better. However, the end user should always have their own choice be valid.

arbiter-engine/src/examples/token_minter.rs
Autoparallel: This is WIP. Hard to know this right way to abstract this at the moment. I'm thinking about it a lot as I work on other feats. Helpful to have other features in there so we can revisit the World set ups with actual simulation on top of them.

PR #801, opened 2024-01-19
feat(arbiter-core / arbiter-engine): broadcasters, halting, test sims

### LINKED ISSUE #756: enhance(arbiter-core): use a broadcast (like `tokio::sync` or `async-broadcast`) in `Environment.Socket`
# Idea
Right now we store a collection of `Sender`/`Receiver` pairs to handle broadcasting. As the number of listeners increases, this becomes more inefficient. With the right implementation, it should also reduce the code complexity a bit as we require a mutex lock to add senders to the current broadcaster.

## Solution
Both the module `tokio::sync::broadcast` and the crate `async_broadcast` provide a MPMC channel system that broadcasts such that all listeners can receive messages in their channel. We can replace the `Arc<Mutex<EventBroadcaster>>` with an `EventBroadcaster` that uses one of the channels discusses prior. We should not need an `Arc<Mutex<..>>` wrapper as receivers can be cloned with no lock needed.

### LINKED ISSUE #788: feat(arbiter-engine): `World` driving `Agent` behaviors concurrently

Autoparallel:
Managed to switch over to using `tokio::sync::broadcast` as the broadcaster of choice in both `arbiter-core` and `arbiter-agent` which seemed to fix concurrency issues with agent behaviors. This gets both the issues below.

- [x] Closes #788 
- [x] Closes #756

** IN PROGRESS**
`World::stop()` is broken, however agents can halt their behavior processing by choice in definition of that behavior.
Replaced with #802


0xJepsen (approved):
I'm into merging this. Nice work fighting this battle with the async demons

PR #804, opened 2024-01-22
feat: `Signer` trait for `RevmMiddleware`

Autoparallel:
Implements the `Signer` trait for the `RevmMiddleware` struct.


0xJepsen (approved):
seems simple enough to me!

PR #820, opened 2024-01-26
refactor(arbiter-engine): agent-behavior flow

### LINKED ISSUE #819: enhance(arbiter-engine): `Behavior`s receive messager and client from above
# Idea
It is cumbersome to have `Behavior`s that `Agent`s own have to get their `Messager` and `Arc<RevmMiddleware>` in their constructor. Instead, we should inherit it from the `Agent` that owns this `Behavior`.

## Solution
We can have the `sync()` stage of the `Behavior` take in the `Messager` and `Arc<RevmMiddleware>`. This can be passsed down using the `MachineInstruction`.

Autoparallel:
This PR adds the ability for the `sync()` stage of a `Behavior` to receive the client and messager from the agent the behavior is attached to. It should then be recommended to users to make these fields optional in their behaviors at some point down the road, but we are not done with this design yet. There could be better ways to address this problem, but I'll take this for now.

Closes #819

### DIFF
--- a/arbiter-engine/src/agent.rs
+++ b/arbiter-engine/src/agent.rs
@@ -144,8 +144,9 @@ impl Agent {
     pub(crate) async fn run(&mut self, instruction: MachineInstruction) {
         let behavior_engines = self.behavior_engines.take().unwrap();
         let behavior_tasks = join_all(behavior_engines.into_iter().map(|mut engine| {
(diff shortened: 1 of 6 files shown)


0xJepsen (approved):
nice this seems like a clear improvement on the prior version.

PR #821, opened 2024-01-27
feat(arbiter-core): console log

### LINKED ISSUE #822: feat(arbiter-core): Console log contract in our revm instance
# Idea
Give the arbiter `revm` instance the ability to deploy the console log contract. 
This console contract is a library. What contract do we use around this?

When you write a test contract you extend it with `std-test` Matt hypothesized this is where the contracts are extended with the is test modifier. 

How can we add this to arbiter in the most idiomatic way. 

Tried deploying a few different bytecodes, own version of console.logs.

Autoparallel:
Closes #822


0xJepsen (dismissed):
This is sick that the inspectors are in here now

kinrezC (approved):

PR #826, opened 2024-01-30
feat(arbiter-engine): run multiple worlds in parallel

### LINKED ISSUE #505: feat: give `Manager` batch runner (and/or config parser?)
# Idea
In current simulations like [`portfolio_simulations`](https://github.com/primitivefinance/portfolio_simulations) we require an end user to not only create their own configs (bad), but to parse them themselves (worse), and moreover to spawn tokio tasks to run parallelized simulations (horrible). 

We should abstract away as much of this as possible. At the very least, a batch runner could be made and the end user can create a bunch of environments but we should provide the concurrent/async logic so a simpler API can be given. Ideally, we can reduce `main.rs` in `portfolio_simulations` drastically. Further, we could conceivably remove `config.rs` in `portfolio_simulations` all together.

Yes, there is far too much code (and it is difficult) for an end user to write!

## Solution
Solution is not clear yet. I will investigate some techniques.

## Alternative approaches
I look forward to hearing ideas!

Autoparallel:
We now have a `Universe` struct which can hold on to a collection of `World`s. It provides a means to run all of these worlds in parallel. More fine tuning and features can be added later.

Closes #505


0xJepsen (approved):
This seems straightforward enough.

Issue #824, opened 2024-01-29
feat(arbiter-core): Block production


0xJepsen:
# Idea

This would be an optional feature as it does incurs a cost as block production is expensive.

PR #863, opened 2024-02-08
feat(arbiter-core): stream with meta

### LINKED ISSUE #501: bug: cannot `stream_with_meta()` on `Event`s
# Problem
We can run `event.stream()` for a given `ethers::contract::Event`. However, `event.stream_with_meta()` fails as we don't populate the necessary data for this. It should be possible to do so.

---

## Steps to reproduce
Create an event and try this like so:
```rust
liquid_exchange.swap_filter().stream_with_meta();
```

## Expected behavior
We retrieve the necessary log metadata.

## Code snippets/links or screenshots 
I was trying this in [`portfolio_simulations`](https://github.com/primitivefinance/portfolio_simulations):
```rust
let events = self.events;
        let mut handles = vec![];
        for event in events {
            handles.push(tokio::spawn(async move {
                println!("Listening for events");
                // let series = Series::new(, _);
                let mut meta_stream = event.stream_with_meta().await.unwrap();
                if let Some(Ok(item)) = meta_stream.next().await {
                    let thing = item.1;
                    println!("Meta: {:?}", thing);
                };
                let mut stream = event.stream().await.unwrap();

                while let Some(Ok(event)) = stream.next().await {
                    println!("Event: {:?}", event);
                }
            }));
        }
```
but get a panic:
```
Listening for events
thread 'tokio-runtime-worker' panicked at 'should have a block number', /Users/colin/.cargo/git/checkouts/ethers-rs-278eab0da531a0a1/5ef227c/ethers-contract/src/log.rs:47:44
```

Autoparallel:
Now we can call `my_contract.events().stream_with_meta()`. Not all fields are populated correctly, and this can be addressed in #824 if we want.

Closes #501

### DIFF
--- a/arbiter-core/Cargo.toml
+++ b/arbiter-core/Cargo.toml
@@ -61,7 +61,6 @@ futures.workspace = true
 cargo_metadata = "0.18.1"
 chrono = "0.4.33"
 
-
 assert_matches = { version = "=1.5" }
 
 [[bench]]
--- a/arbiter-core/src/data_collection.rs
+++ b/arbiter-core/src/data_collection.rs
@@ -310,9 +310,9 @@ impl EventLogger {
                         }
                         break;
                     }
-                    Broadcast::Event(event) => {
+                    Broadcast::Event(event, receipt_data) => {
                         trace!("`EventLogger` received an event");
-                        let ethers_logs = revm_logs_to_ethers_logs(event);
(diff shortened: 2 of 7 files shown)


0xJepsen (approved):
This looks good, i does seem to break one test. It looks like it should be easy to propagate the change to the test. Let me know if you want me to take care of it.

PR #864, opened 2024-02-08
migrate to anyhow::Result

Autoparallel:
Some of the traits and their return types are not easily managed with custom errors, so they are replaced with `anyhow::Result` to make the end user's life easier.


kinrezC (approved):
LGTM

PR #868, opened 2024-02-09
nit: remove `PhantomData`

ts0yu:
removes redundant `PhantomData` in `Engine`

was originally used for type parameter `E` but now `event_stream` utilizes `E` so `PhantomData` is redundant


Autoparallel (approved):
Easy and clean enough! Thanks!

PR #889, opened 2024-02-14
Rename Directories

kinrezC:
- Renames crate directories from `arbiter-*` to `*`, e.g. `arbiter-core` -> `core`.  Crate names are still `arbiter-*`.


Autoparallel (changes requested):
It looks like the book has an issue (check the book.toml as it is using this to build and since the dir was renamed to `docs/` this is borked).

Also, there are some still git submodules using `arbiter-bindings/lib`, can we get those changed too?

Lastly, while we're here, we should probably have the template `examples/arbiter-template/` just be `examples/template/`

Autoparallel (approved):
lgtm!

0xJepsen (approved):

PR #898, opened 2024-02-16
feat(core): let clients use coprocessor

### LINKED ISSUE #838: enhance(arbiter-engine): allow `Behavior`s to get their own `Coprocessor`
# Idea
The `Coprocessor` is ready to be tested and used. What better use case is there than giving all of your behaviors a way to have their own scratch pad that won't block the main tx stream? This is the way!

### LINKED ISSUE #853: chore(`ethereum`): coprocessor should have a configurable inspector
# Idea
Currently the co-processor is not configured with any REVM inspectors. 

## desired behavior
We would like for it to inherent the inspectors from environment but subtract the gas (processors don't need to pay gas)

Autoparallel:
So far a bit of a dependency cleanup and now the `Connection` has a `Coprocessor` and I had to `unsafe impl Send + Sync` on the `Coprocessor` due to some of the handler `dyn Fn`s not being `Send + Sync`. This now needs testing and integration.
- [ ] Use the `Coprocessor` for `.call()` method in `ArbiterMiddleware`.
- [ ] Test the `Coprocessor` maintains state and does not have any glaring bugs due to `Send + Sync` unsafes.

- [x] Closes #853 
- [ ] Closes #838 (by virtue of having this straight in the client)


Autoparallel:
This is actually very problematic as the EVM cannot be `Send` and `Sync` without a wrapper and unsafe override. Ditching for now.

PR #899, opened 2024-02-16
feat: make streams optional for behaviors for oneshot

### LINKED ISSUE #897: feat(arbiter-engine): optional stream and default process
# Idea
It's possible some behaviors are just one-shot and don't ever need to process. We should allow for this by returning an optional stream and having a default implementation of process.

Autoparallel:
To make "oneshot" behaviors easier, we can just have `startup()` return an optional stream and also have a default impl for the `process()` method. 

Closes #897


kinrezC (approved):
LGTM

PR #911, opened 2024-02-25
feat: environment logs

### LINKED ISSUE #869: feat: Implement `get_logs()`
tbd more information

Autoparallel:
# Work Done
In this PR, I update the `ArbiterDB` to also store a `HashMap` of block numbers to logs emitted in that block. As a bonus, we:
- [x] Close #869 

--- 
**OLD STUFF HERE**
# Idea
What I'm trying to do here is allow the `Environment` to store a `global_db: BTreeMap<U256, CacheDB>` so that each KV pair in the `global_db` will be a representation of the state at the end of the block, including logs that were emitted.

The goal here is to create a workable db organized by blocks so that (middleware or otherwise) requests that can take in a `block_number` can be satisfied by lookup in this db. Furthermore, if we can take this `global_db` and output it into hard storage, we can use this for extracting data from and analysis with. Tools like `cryo` become more of an option for us, and we could conceivably avoid having to use the `events::Logger`. 

## Progress
- [x] Added 3loc to the `ArbiterInspector` that will collect logs into the evm.
- [x] Added logic in the `Instruction::BlockUpdate` that will add the current `CacheDB` into the `global_db` and clear the logs.

## Questions
- There is a tradeoff here of storing the whole state each time versus storing diffs. Storing the whole state can cause bloat in memory and is in some sense unnecessary. However, computing diffs could be inefficient in speed and would cost more in reconstruction of state later on.
    - What is the right choice? 
    - Do we provide optionality here?
    - Can we look towards `reth` for inspiration on how to handle this nicely?
- Can we offload some of this additional complexity with concurrency/parallelism? The `events::Logger` uses its own thread to do work. Perhaps we could do something of the same sort.
- What will benchmarks say here? We definitely need to do some and compare against Anvil yet again. 
- If we don't want to store the whole chain state for any of the above reasons or otherwise, we can certainly just store logs which is a great start! 
    - What is the minimal set of items to store to give excellent end-user analysis tooling? Sometimes having the right things emitted as logs is tough.
    - What ways can we collect the "right set of items" that could be ergonomic and efficient? For instance, perhaps we can set up inspectors in such a way to monitor specific contract state if that is really our goal. Inspectors can monitor if certain storage locations are hit for sure, but this could be tough to implement properly.
    
---

Let's please discuss this. I think that something along these lines and leading into an `arbiter-analysis` crate that uses Rust-eth tools like `cryo` could reduce complexity in using the data output by Arbiter if done correctly.

### DIFF
--- a/core/src/environment/mod.rs
+++ b/core/src/environment/mod.rs
@@ -569,7 +569,7 @@ impl Environment {
                                 let to_block = U256::from(
                                     filter
                                         .block_option
-                                        .get_from_block()
+                                        .get_to_block()
                                         .ok_or(ArbiterCoreError::MissingDataError)?
                                         .as_number()
                                         .ok_or(ArbiterCoreError::MissingDataError)?
--- a/core/src/environment/mod.rs
+++ b/core/src/environment/mod.rs
@@ -258,7 +258,10 @@ impl Environment {
                     label
                 );
                 match instruction {
-                    Instruction::AddAccount { address, .. } => {
+                    Instruction::AddAccount {
+                        address,
+                        outcome_sender,
+                    } => {
                         let recast_address = Address::from(address.as_fixed_bytes());
                         let account = revm::db::DbAccount {
                             info: AccountInfo::default(),
--- a/core/src/environment/mod.rs
+++ b/core/src/environment/mod.rs
@@ -72,7 +72,9 @@ pub struct Environment {
 
     /// The [`EVM`] that is used as an execution environment and database for
     /// calls and transactions.
-    pub(crate) db: ArbiterDB,
+    pub(crate) tip_db: ArbiterDB,
+
+    pub global_db: Arc<RwLock<BTreeMap<U256, CacheDB<EmptyDB>>>>,
 
     inspector: Option<ArbiterInspector>,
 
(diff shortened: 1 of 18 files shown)


0xJepsen:
I am very curious what the speed implications are for this. I agree that benchmarks are important.

Autoparallel:
The least expensive would be to just store the logs which shouldn't be too bad at all.

core/src/environment/mod.rs
```diff
@@ -72,7 +72,9 @@ pub struct Environment {
 
     /// The [`EVM`] that is used as an execution environment and database for
     /// calls and transactions.
-    pub(crate) db: ArbiterDB,
+    pub(crate) tip_db: ArbiterDB,
```
0xJepsen, line 75: Whats the intuition of calling it tip? Like it's the front of a queue? Or top of a stack?

core/tests/environment_integration.rs
```diff
@@ -148,3 +148,39 @@ async fn env_returns_db() {
...
+        .send()
+        .await
+        .unwrap()
+        .await
+        .unwrap();
+
+    println!("Global DB: {:?}\n", global_db.read().unwrap());
+
+    // UPDATE BLOCK
+    let new_block_number = 69;
```
0xJepsen, line 169: 🌝

0xJepsen (approved):
this seems pretty clear and straightforward to me. I think this is a good step towards the right abstraction of blocks

PR #913, opened 2024-02-25
feat: `error!` output from behavior

### LINKED ISSUE #912: fix(engine): error propagation
# Idea
Errors that happen inside of either the `startup` or `process` methods of behaviors are not properly bubbled up and can be very hard to see. We should fix this.

## Possible Solution
We can just log an `error!` for now which will at least output this when monitored with `tracing_subscriber`. This is a quick and dirty approach that won't break anything.

Autoparallel:
Closes #912 

We could likely improve this further, but this was a quick fix that should at least let errors be visible.


0xJepsen:
looks like this changes the expected return type in the token requester test from an `option<box<pin<stream>>>` to just the `box<pin<stream>>`

PR #922, opened 2024-02-29
core: db management

Autoparallel:
**WIP**

Adding methods here to make a `World::builder()` with some nice configuration as well as allow us to get the `ArbiterDB` we get from the `Environment::stop()`


Autoparallel:
I'm rebasing off the #911

kinrezC (dismissed):

0xJepsen (approved):
lgtm, the builder pattern is op

PR #977, opened 2024-05-24
feat: Expose underlying cache_db with env.db() function

jbargu:
**Give an overview of the tasks completed**
Expose underlying cache_db with `env.db()` function. This is useful to check the execution state and environment re-creation without stopping the simulation.

**Link to issue(s) that this PR closes**

**REMINDER! Please check that you have done the following prior to submitting this PR:**
- [x] Checked that the relevant version(s) have been incremented if necessary.
     - please advise if version is to be incremented
- [x] Ran both and made any changes necessary for:
    - `cargo +nightly fmt --all`
    - `cargo clippy --all`
    
You can install the nightly toolchain on your system by:
```bash
rustup toolchain install nightly
```

PR #980, opened 2024-05-31
feat: Add Cheatcodes::AddAccount

jbargu:
Arbiter currently assumes all accounts are initialized before the execution. However sometimes the user cannot anticipate all the involved accounts. With the `Cheatcodes::AddAccount` cheat code, the user can inject a new account mid execution and then update the balance with `Cheatcodes::Deal` cheat code.

The PR is based on top of #977.

**REMINDER! Please check that you have done the following prior to submitting this PR:**
- [x] Checked that the relevant version(s) have been incremented if necessary.
- [x] Ran both and made any changes necessary for:
    - `cargo +nightly fmt --all`
    - `cargo clippy --all`

### DIFF
--- a/core/src/environment/instruction.rs
+++ b/core/src/environment/instruction.rs
@@ -223,6 +223,12 @@ pub enum Cheatcodes {
         /// The address of the account to fetch.
         address: ethers::types::Address,
     },
+    /// An `AddAccount` is used to add a default/unfunded account.
+    /// If the account exists, the value is updated.
+    AddAccount {
+        /// The address of the account to be added.
+        address: eAddress,
+    },
 }
 
 /// Wrapper around [`AccountState`] that can be serialized and deserialized.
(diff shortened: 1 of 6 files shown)


Alexangelj:
Reviewing next week

Alexangelj:
Overall like the changes - just need to make sure everything merges in clean

jbargu:
At first I wanted to expose the whole `Instruction` struct and add `Cheatcodes(Instruction)` but then the struct is cyclic. 

Maybe a more sound way would be to have an `apply_instruction(Instruction)` function or modify `apply_cheatcode(Cheatcodes)` to `apply_cheatcode(Instruction)` however this would be a breaking change.

Alexangelj:
Actually - maybe we can just keep the existing apply_cheatcode and expose a new method `apply_instruction(Instruction)`, @kinrezC wdyt?

PR #1003, opened 2025-05-29
feat(`core`): rework

Autoparallel:


arbiter-core/src/runtime/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+  // Example agent types - simple structs with clear state
+  #[derive(Clone)]
+  struct Counter {
+    total: i32,
+  }
+
+  impl LifeCycle for Counter {}
+
+  #[derive(Clone)]
+  struct Logger {
```
ConnorMJones, line 476: The interface here seems like it's working well. Fairly simple.

arbiter-core/src/handler.rs
```diff
@@ -0,0 +1,35 @@
...
+  type Reply: Message;
+
+  fn handle(&mut self, message: M) -> Self::Reply;
+}
+
+// Simplified handler function type - no more complex wrapper
+pub type MessageHandlerFn = Box<dyn Fn(&mut dyn Any, &dyn Any) -> Rc<dyn Any>>;
+
+/// Create a handler function for a specific message type
+pub fn create_handler<A, M>() -> MessageHandlerFn
```
ConnorMJones, line 19: Seems as good as it can be for now. I don't think you can get around the Any here or this design without changing how the Agent + handles work. The other way that you might be able to get this to work is have the message and response types on the agent and the handle derived from the agent.

ConnorMJones (approved):

